PR #2, opened 2010-10-11
OSGi-fied slf4j-api by means of Maven Bundle plugin (BND)

hseeberger:
As discussed on the mailing list, please "carefully inspect" the created manifest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
           <configuration>
             <instructions>
               <Bundle-RequiredExecutionEnvironment>J2SE-1.4</Bundle-RequiredExecutionEnvironment>
-              <_versionpolicy>[$(@),$(version;=+;$(@)))</_versionpolicy>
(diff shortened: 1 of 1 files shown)


mattbishop:
Heiko is this PR still valid? I am using many slf4j libs in an OSGi project and the manifests seem to be working just fine for me.

As the PR is quite old now I wanted to see about closing it if it's not needed any more.

immetoo:
@mattbishop Yes slf4j does not work in OSGi bundles, it can't be loaded when deploying one jar a time.
Because it can't deploy logback as it deps on slf4j and can't deploy slf4j as it depends on impl which waits on slf4j.

After many hours have found the workaround, just add one fake class in the correct package to boot and disable logging.
`
package org.slf4j.impl;

public class MyClass {
	
	public static void fakeDepToBootSLF4J(String out) {
	}
}

`



btw: your site gives a blank default Apache Tomcat/9.0.68 page

PR #4, opened 2010-11-03
Relaxed versioning policy as discussed on the mailing list

hseeberger:


mattbishop:
Heiko, same comments as #4

PR #8, opened 2011-11-05
Remove unnecessary interface modifiers.

motlin:


motlin:
- Modifiers 'public' and 'abstract' are redundant for interface methods.
- Modifiers 'public', 'static', and 'final' are redundant for interface fields.

ceki:
Redundant perhaps but not harmful either. While I do not wish to be discouraging, I don't see the added value of this PR, and accepting it for the sake of encouragement seems dishonest.

PR #28, opened 2012-10-25
Misc slf4j-simple fixes and enhancements

javabean:
- Java 5 syntax
- add SimpleLoggerAccess class to enable to change logger levels at runtime
- update documentation (including example simplelogger.properties file)
- restore ability to completely turn off logging for a given logger (level "off")
- force output file (if any) to UTF-8 to avoid platform-dependent surprises while logging to a file


jgangemi:
any chance adding 'off' back to `SimpleLogger` could be pulled into it's own request? i'd find that option very useful to turn off logging in some of my unit tests.

PR #43, opened 2013-06-29
Fixed situation where migrator does not convert JCL log statements and breaks code.

steve217:
There is a bug in the migrator: when the migrator encounters the method line on a new line apart from the class, the convertor changes the declaration but not the method name and thus breaks the code.

Normal convert:

``` java
Log logger = LogFactory.getLog(MyClass.class);
```

To

``` java
Logger logger = LoggerFactory.getLogger(MyClass.class);
```

When method name is on a separate line...

``` java
static Log logger = LogFactory
            .getLog(DoesntWork.class);
```

.. the migrator changes the declaration ... but not the type and method name.

``` java
static Logger logger = LogFactory
    .getLog(DoesntWork.class);
//sb  LoggerFactory
//    .getLogger(DoesntWork.class);
```

---

This looks like an old bug.
See http://mailman.qos.ch/pipermail/slf4j-dev/2011-March/003280.html

---

Also I fixed the spelling of 'instanciation'. :)

### DIFF
--- a/slf4j-migrator/src/main/java/org/slf4j/migrator/line/JCLRuleSet.java
+++ b/slf4j-migrator/src/main/java/org/slf4j/migrator/line/JCLRuleSet.java
@@ -64,6 +64,13 @@ public class JCLRuleSet implements RuleSet {
         .compile("LogFactory.getLog\\("),"LoggerFactory.getLogger(");
     
 
+    SingleConversionRule cr6 = new SingleConversionRule(Pattern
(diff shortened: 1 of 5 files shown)


steve217:
Matt - I changed the rule I used to a MultiGroupConversionRule version,, but I honestly couldn't figure out how to use that class.(I see in TrivialMatcher (46) where the class is employed) but is the point of that class to provide multiple patterns for a single replacement?

PR #29, opened 2012-12-09
Implementation of LoggerFactory.getLogger() which determines the calling class name through reflection.

adorokhine:
Allows you to change invocations such as
    LoggerFactory.getLogger(MyClass.class);
to
    LoggerFactory.getLogger();

This is a convenience method for simpler Logger creation, as well as allowing you to copy and paste Logger creation code without having to remember to change the classname.

PR #44, opened 2013-07-11
A way to get a default logger for the current class.

ggerard:
I had this in a LogUtility class that I've been using for a while and think it would be handy for others.


mattbishop:
This feature has been addressed in #29 so I am closing this PR. Thanks for contributing, I hope you enjoy using the simpler getLogger().

PR #45, opened 2013-07-27
Added minimal implementation of log4j's class SimpleLayout.

reflix:
Hi!

I recently faced a ClassNotFoundException as the class SimpleLayout is not provided by the log4j binding (log4j-over-slf4j).

This adds a minimal implementation of the SimpleLayout to the log4j-over-slf4j module.

Regards,
- Felix


epabst:
This appears to already be done, hence the conflict.  I recommend closing this pull request.

PR #57, opened 2014-01-29
SLF4JBridgeHandler.Initializer

mfulton26:


mfulton26:
Refreshed. Thanks.

jul-to-slf4j/src/main/java/org/slf4j/bridge/SLF4JBridgeHandler.java
```diff
@@ -300,4 +313,26 @@ public void publish(LogRecord record) {
...
+     */
+    public static class Initializer {
+
+        /**
+         * Registers {@link SLF4JBridgeHandler} as handler for the j.u.l. root logger.
+         *
+         * @throws IOException if there are problems reading from the stream.
+         */
+        public Initializer() throws IOException {
+            String loggingConfigurationString = "handlers = " + SLF4JBridgeHandler.class.getName();
```
mfulton26: I've added a test locally and now I realize that I don't need to create an in-memory configuration file to read. I will push an update in a bit.

mfulton26:
Unit test added and main code simplified. Please let me know if any changes are needed. This should be ready for the next release. Thanks.

jul-to-slf4j/src/test/java/org/slf4j/bridge/SLF4JBridgeHandlerTest.java
```diff
@@ -178,4 +179,10 @@ void assertLevel(int index, org.apache.log4j.Level expectedLevel) {
         LoggingEvent le = (LoggingEvent) listAppender.list.get(index);
         assertEquals(expectedLevel, le.getLevel());
     }
+
+    public void testInitializer() throws IOException {
```
mfulton26: This no longer needs `throws`...

mfulton26: Updated and ready for review.

mfulton26:
I had to add back logic not using `LogManager.getLogManager().getLogger("")` because the root logger, "", is not yet initialized when a j.u.l. configuration class is constructed (it causes a `NullPointerException`).

Ready for review.

anthavio:
Why oh why is this still not merged in 2020 so I have to copy this again...

mfulton26:
@ceki, any concerns with merging this?

@anthavio I don't actively work in Java these days but if you or someone else would be willing to update this PR (I see the unit test file has some conflict(s) now) I am confident that @ceki would be willing to approve and merge.

anthavio:
OK, I will try to refresh this PR to mergeable state

ceki:
Please create a Jira issue for this PR.

anthavio:
https://jira.qos.ch/browse/SLF4J-489

PR #70, opened 2014-04-23
Add jdk 1.8 suppport 

ouertani:
Call by name (or lazy evaluation)  will be accessible using Supplier of String. The example below : 

```
 logger.trace(()  -> "trace msg "+echo()) ;
```

echo method and String concatenation will only be evaluation if trace level is enabled.


dfa1:
GJ

ceki:
SLF4J needs to support JDK 1.5 whereas the Supplier interface was added in JDK 1.8. Even if the Supplier interface was available in JDK 1.5, we can't modify the Logger interface in order to preserve backward binary compatibility.

ouertani:
@ceki  the jdk 1.8 support has been introduced as a separate module like jdk 1.4 support and not part of current and default api interfaces. It is import to note that introducing this module does not break existing libraries and interfaces as new methods has been add as default implementation.

ceki:
I see. What would happen if some application A depended on two modules say B and C with module B depending on slf4j-api.jar and module C depending on slf4j-jdk18.jar? To make the problem easier, further assume that  the user managed to include only slf4j-jdk18.jar in her class path (no  slf4j-api.jar). Can classes in module B be loaded and run OK?

I personally cannot answer the above question with certainty. Can you?

ouertani:
@ceki  what you describe here is a general problem that can be raised even with different version of the same libraries.  As said, this module don't break compatibility, and one of the sample solution is to exclude the default slf4j-api because it is fully covered by slf4j-jdk18.jar

emmanueltouzery:
That does not solve the main problem (frozen Logger interface), but [according to java documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html):

> However, the compiler will treat any interface meeting the definition of a functional interface as a functional interface regardless of whether or not a FunctionalInterface annotation is present on the interface declaration.

In other words I'm basically certain that if you were to write:

```
public interface StringSupplier {
   public String get();
}
```

That code would compile with java5 and would enable you to type lambdas in a java8 client.

But because of the `Logger` interface freeze that's not really helpful sadly.

emmanueltouzery:
I would also add that if a jdk-1.8 interface is added, that could solve to a degree future problems with extensibility of the `Logger` API thanks to the new java8 feature of default methods in interfaces. That would enable adding new methods to the interface in the future without breaking compatibility.

ouertani:
@emmanueltouzery the main feature that has been add ( as new module like jdk-1.4 support) is the default non-intrusive supplier methods params.

emmanueltouzery:
@ouertani of course, and that's the reason why I'm interested in this bug. I'd like the feature to be merged, whether it's implemented through a jdk-1.8 module or through some extension of the existing 1.5 code.

smtibaa:
Yes !! thanks Slim

BenRhouma:
Nice job @ouertani , works like a charm, I really enjoy the lazy evaluation feature, that's why I have merged your PR in my repo and switched to slf4j-api-jdk18, https://github.com/BenRhouma/slf4j

the lazy evaluation is really helpful to enhance performance when running in prod, especially when we use an algorithm full of trace or debug logs. 

###### Result using a simple test:
``` java
  Logger.info  ("info " + echo());
  Logger.trace("trace"+ echo());

public static  String echo (){
     Logger.info("echo invoked");
     return "echo body";
}
 ``` 
    
#### Main difference between SLF4j and SLF4j-JDK18

##### output using slf4j-api and trace is disabled
17:16:56.403 [main] INFO  - echo invoked
17:16:56.406 [main] INFO  - info echo body
17:16:56.406 [main] INFO  - echo invoked <-- **method echo is invoked even if trace is disabled**

##### output using slf4j-jdk18 and trace is disabled

17:16:56.403 [main] INFO  - echo invoked
17:16:56.406 [main] INFO  - info echo body

**Comment by azaroui** [comment]
+1 This is very valuable functionality. Would love to see this in master. Thanks @ouertani!

**Comment by napilioni** [comment]
Indeed very good functionality. +1

**Comment by nathansgreen** [comment]
This doesn't appear to support parameterized strings, which is pretty limiting. It's also very difficult to tell what's actually different since the an entire module was copied, resulting in a lot of duplication.

**Comment by garagatyi** [comment]
@nathansgreen I saw you [proposal](http://mailman.qos.ch/pipermail/slf4j-dev/2015-October/004231.html) of lambda support in slf4j mailing list. Can you explain what are the pros of your solution? Maybe you can submit PR with your solution so author of current PR might close it if your solution is better or has better feedback from slf4j maintainers.

**Comment by nathansgreen** [comment]
@garagatyi I was hoping for some feedback/constructive criticism before opening a PR. I don't know what features people really want with a lambda-enabled API, so the PR history could get pretty convoluted. I also have two separate branches with two different implementations. Either could be used independently or they could be merged, but I won't know what needs to happen just thinking about it all by myself. Also, I haven't had time yet, but my plan was to convert a bunch of code from Spring Framework to the new API to get a feel for how it looks in a decent-sized project with decent code quality.

Pros:
1. No code is copied. The existing PR has a lot of copied code and I think that's unnecessary.
2. The regression test suite provides strong confidence in binary compatibility. SLF4J has long emphasized compatibility, so I don't feel the need to argue for its value.
3. It's easy to use for most common cases. I rarely need to make more than two expensive calls for a parameterized logging statement, so that's covered.
4. Correctness. The following example is possible and I've seen it in the wild a handful of times:

```java
if (logger.isInfoEnabled()) {
    logger.debug("Message {} with {} really {} expensive {} arguments, a.eval(), b.eval(), c.eval(), d.eval());
}
```

If lambdas are first-class in the logging API, this problem goes away. (It ought to be caught in code review, but mistakes happen.)

The other thing I was hoping to put a bit of time on was more low-level analysis. Performance testing and digging into the compiler-generated bytecodes would provide the data needed to document the new API and when/if the old API is more appropriate.

garagatyi:
Thank you for fast response. 
I agree that backward compatible code will be much more valuable.
I think backward compatibility was the main reason why @ceki didn't put much attention to that PR.
Did you receive any response from slf4j maintainers?

ceki:
Indeed, a new module replicating the contents of slf4j-api is not going to fly, whereas Nathan Green's approach which ensures compatibility is quite appealing. 

In addition to lamdas I would like to initiate an effort for making the Logger interface more composable (instead of adding more methods). I'll post something tomorrow on the slf4j-dev mailing list with more details.

ouertani:
@garagatyi @ceki the api is full compatible with the default api ( by the default methods).

ceki:
The [javadoc of @FunctionalInterface](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html) states
> However, the compiler will treat any interface meeting the definition of a functional interface as a functional interface regardless of whether or not a `@FunctionalInterface` annotation is present on the interface declaration.

This means that by defining our own functional interfaces, e.g. `Supplier`, and adding methods accepting `Supplier` as parameter in org.slf4j.Logger, SLF4J can support lambdas even if compiled under JDK 1.5 as long as the client uses JDK 1.8.

In any case, having a separate module for supporting lamdas, i.e. slf4j-jdk18, as proposed in this PR is unacceptable.

MuhammadHewedy:
We need this feature, at least as @emmanueltouzery suggests

MuhammadHewedy:
Noted that it is already part of JUL, and this way slf4j hiddes an important feature from one of the apis it wrapps

MuhammadHewedy:
Why not make like spring do, build with java8 with -source and -target 1.6 (or whatever version you support), and do runtime checks like (if class present) before use java8 feature. https://spring.io/blog/2015/04/03/how-spring-achieves-compatibility-with-java-6-7-and-8

nathansgreen:
The short answer is that Spring is a container and manages classloading, so
it can do bytecode detection and generation and other cool things. SLF4J
doesn't do this, and therefore cannot rewrite uses of the Logger interface,
which is the primary abstraction everyone works with.

rkraneis:
I don't see how the two diametrically opposed goals of "keeping the api untouched" and not "using java8 functionality with two different artifacts" could be achieved.

I see three (or four) choices:

1. Have, as proposed by @ouertani, two different artifacts - at least for the core changes involving lamdas. One could factor out the remaining parts into a third shared artifact. This would allow everyone on Java 8 to use the new and improved API while maintaining compatibility with loggers implementing the old interface.
    1. Do the differentiation by name or classifier. This could lead to two competing slf4j-apis on the classpath but can be easily resolved via maven excludes.

    2. Bump the new API to e.g. slf4j-2.x and keep backports (if needed) to it in a slf4j-1.x branch. This would nicely remove the dependency problem and still allow internal updates.

2. Keep one artifact and extend the ILogger interface (and introduce a StringSupplier) - losing compatibility with Loggers implementing the old interface. Using default methods forces the API user on Java 8.

3. Introduce a new ILambdaLogger and a StringSupplier keeping the ILogger Interface intact and everything on Java 1.5.

rkraneis:
To answer your question @ceki:

> What would happen if some application A depended on two modules say B and C with module B depending on slf4j-api.jar and module C depending on slf4j-jdk18.jar? 

Maven would include both apis leading to an unstable build. The older api must be excluded manually. The same problem exists with libraries depending on loggers for which slf4j supplies bridges. To detect these you can use the [banndedDependencies Rule](https://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html)

> To make the problem easier, further assume that the user managed to include only slf4j-jdk18.jar in her class path (no slf4j-api.jar). Can classes in module B be loaded and run OK?

Yes, this is the core purpose of default method. As an example Javas [Iterable](https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#spliterator--) was extended that way to also support Spliterators.

nathansgreen:
René,

1.2 is essentially what I was proposing: Bump the version (to at least 1.8)
and retain binary compatibility. Changes to the Logger interface are all
done with default methods so that any existing implementation will continue
to work. Clients running under JDK 6 or 7 would deploy the 1.7.x API jar
and not have to deal with special exclusions or classifiers. If a jar
compiled against 1.8.x needs to be included, it can be rewritten using a
tool like Retrolambda.

Nathan

ceki:
API improvements along with a review process are planned for SLF4J 1.8.

JasonMing:
An single method interface (without `@FunctionalInterface`) could be defined in the slf4j-api artifact to resolve the backward compatibility.
Do not using the `Supplier` interface from the JDK1.8, using an interface like following:
```java
public interface LogMessageSupplier
{
    String get();
}
```
The 1.8+ compiler will accept the lambda expression when passing the argument to the `LogMessageSupplier` parameter. Nothing will be broken.

ouertani:
@JasonMing  this is yet another guava and as LogMessageSupplier didn't implement the Supplier interface, a wrapper layer will be a inevitable to reuse java 8 interfaces.

JasonMing:
@ouertani Is that really necessary to reuse java 8 interface? Every single method interface accepts the lambda expression, the jdk1.8 will do this for you and with no breaks for the previous jdk.

fzakaria:
I'd like to see this pulled through.
IMO : as the change is currently in a separate module, re-using the JDK8 interfaces should be preferred.

eduardsi:
Any news on lazy evaluation support in the API, folks?

xenoterracide:
even [JUL has this](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html#info-java.util.function.Supplier-)

nhojpatrick:
any update?
any help needed?

nathansgreen:
@nhojpatrick this was discussed on the mailing list a while ago but then stalled out. I don't see any future for this PR: it seems unlikely to ever be merged and should probably be closed.

ouertani:
@nathansgreen Why this PR should be closed? It does not impact any core modules.

ceki:
A new slf4j-api module is out of the question.

nhojpatrick:
Instead of a new module what about the next version bumps it's minimum version to Java 8?
As I'm guessing with Java 9, it will mean slf4j will need to do some changes so it works with modules.

hazendaz:
Time for community fork?

fzakaria:
I'm surprised by the attitude here.
The PR was a forward/backwards compatible.

Is there a better outreach to the SLF4J community like a mailing list to take this discussion ?

Would be great to get JDK8+ features.

ceki:
The separate module approach (slf4j-api-jdk18) would be a nightmare for both the maintainers of SLF4J **and** for its users. 

1) Any bug fix in slf4j-api would need to be ported to slf4j-api-jdk18 and vice versa. 

2) Even assuming that slf4j-api-jdk18 is binary compatible with slf4j-api, users would need to exclude slf4j-api and replace it with slf4j-api-jdk18 in their build files. We just can't impose such a change on millions of unsuspecting users.  

As I have indicated repeatedly, while lambda methods in the Logger interface are desirable, in my opinion, a separate module such as slf4j-api-jdk18 is not the way to  get there.

ceki:
@ouertani Is the separate module approach motivated by not wishing to impose Java 8 on users allowing them to stick with slf4j-api which requires only Java 5?

ceki:
This is from  http://mailman.qos.ch/pipermail/slf4j-dev/2016-December/004593.html

Hello all,

Given that JDK 7 earlier are used by over 50% of Java projects, I do not 
think we can force users to upgrade to Java 8. Thus, SLF4J is unlikely 
to bump its JDK requirement to version 8 for yet some time.

On the bright side, there is a way for org.slf4j.Logger to support Java8 
lambdas, or just lambdas, without requiring Java 8. As stated in [1], 
the compiler will treat any interface meeting the definition of a 
functional interface as a functional interface regardless of whether or 
not a @FunctionalInterface annotation is present on the interface 
declaration.

This means that we can create a Supplier interface such as

public interface Supplier<T> {
     T get();
}

and then add the following method to org.slf4j.Logger interface.

package org.slf4j;

public interface Logger {
   ....
   public void debug(Supplier<T> supplier);
   ....
}

This will add lambda support without requiring Java 8 support at build time.

The downside of this approach is that implementations of the 
org.slf4j.Logger interface as found in logback and modules such as 
slf4j-simple, slf4j-log412, slf4j-jdk14 will need to be updated. Given 
that the number of such implementations is limited, I think the 
requirement is quite acceptable.

Your comments are welcome,

nhojpatrick:
As a user, I've got Java 6 projects that already have logging are having
only minor functionality added so I'm happy to stick with the current
version of slf4j being used. Any major functionality changes to a Java 6
project would result in it being upgraded to Java 8.

I've got no Java 7 projects.

I've got Java 8 projects, everything new is Java 8 and older Java 6
projects would be uplifted if major changes happened. So happy to update
slf4j to take advantage of functionality so don't have to do isDebugEnabled
style checks or log messages have delayed construction till actually
required.

From my point of view I don't mind if it's a new module, new
Interface/Supplied, version bump and minimum version bumped. I just feel
slf4j is starting to lag behind other frameworks as Java 8 has been out for
several years and lamdba would be useful for performance reasons within a
logging framework.

John




John

JanecekPetr:
Ceki's proposal sounds solid. Looks like a good solution for everyone.
The downside of adding quite a few new methods to an interface is obviously bad, but could potentially be managed. I guess that means incrementing the major version of slf4j to 2.0.0, right?

xenoterracide:
worth noting that I think that suppliers should work with message format

```
log.debug("foo {} bar {}", foo::toString, () -> foo.getBar().toString );
```

and I've noticed implementing this wrapper myself that java doesn't really (the compiler warns lots...) like a varargs of suppliers, so I ended up making an interface is so far

https://bitbucket.org/xenworks/logging/src/9d1bb594f790ae4bc77e3e7ecd994c0026e12ab4/src/main/java/com/xenoterracide/logging/Logger.java?at=master&fileviewer=file-view-default

I will probably end up licensing this under the apache 2, haven't made it official though as I'm debating on certain aspects of said license. Also it does need some improvements. feedback on what I've done, patches, etc, welcome. If someone wants me to add it to maven central I can probably do that.

BalRoger:
ceki's proposal (https://github.com/qos-ch/slf4j/pull/70?_pjax=%23js-repo-pjax-container#issuecomment-268815061) sounds eminently workable, but I would like to tweak it a bit. Basing the lambda-based logging on the ```Supplier``` interface he defines can cause compilation problems. To wit, if any operation inside the lambda declares a checked exception, such as:

``` java
log.debug(() -> foo.getBar())
```

where

``` java
class Foo {
    ...
    Bar getBar() throws IOException {
        return barGetter.retrieveFor(this);
    }
    ...
}
```

Then the code will fail to compile, since ```Supplier.get()``` does not throw ```IOException```. 

It turns out that there is an easy fix, but it has to be built into the design though. First, redefine the ```Supplier``` interface as follows:

``` java
public interface Supplier<T, E extends Exception> {
     T get() throws E;
}
```

Then, inside the ```Logger``` interface, define each method that uses supplier as follows:

``` java
public interface Logger {
    <E extends Exception> void trace(Supplier<String, E> message) throws E;
    <E extends Exception> void debug(Supplier<String, E> message) throws E;
    <E extends Exception> void info(Supplier<String, E> message) throws E;
    <E extends Exception> void warn(Supplier<String, E> message) throws E;
    <E extends Exception> void error(Supplier<String, E> message) throws E;
}
```

Then the design will be robust enough to withstand checked exceptions inside the lambda.

seanf:
Nice one @BalRoger. Of course the old trace/debug/info/warn/error methods won't declare the new (generic) exceptions, so existing client code won't be affected.

In case anyone else needs to assure themselves that it will compile without casts:

```java
import java.io.IOException;

public class LambdaTest {

    public interface Supplier<T, E extends Exception> {
        T get() throws E;
    }

    public interface Logger {
        <E extends Exception> void trace(Supplier<String, E> message) throws E;
        <E extends Exception> void debug(Supplier<String, E> message) throws E;
        <E extends Exception> void info(Supplier<String, E> message) throws E;
        <E extends Exception> void warn(Supplier<String, E> message) throws E;
        <E extends Exception> void error(Supplier<String, E> message) throws E;
    }

    public static void main(String[] args)
            throws InterruptedException, IOException {
        Logger log = new Logger() {
            @Override
            public <E extends Exception> void trace(Supplier<String, E> message)
                    throws E {
                System.out.println("TRACE: " + message.get());
            }

            @Override
            public <E extends Exception> void debug(Supplier<String, E> message)
                    throws E {
                System.out.println("DEBUG: " + message.get());
            }

            @Override
            public <E extends Exception> void info(Supplier<String, E> message)
                    throws E {
                System.out.println("INFO: " + message.get());
            }

            @Override
            public <E extends Exception> void warn(Supplier<String, E> message)
                    throws E {
                System.out.println("WARN: " + message.get());
            }

            @Override
            public <E extends Exception> void error(Supplier<String, E> message)
                    throws E {
                System.out.println("ERROR: " + message.get());
            }
        };
        log.debug(LambdaTest::debugMessage);
        log.info(() -> {
            Thread.sleep(0);
            return "This lambda can throw InterruptedException.";
        });
    }

    private static String debugMessage() throws IOException {
        return "This method declares IOException.";
    }

}
```

seanf:
FWIW, such a solution would require ugly explicit types if calling from Kotlin:

```kotlin
import LambdaTest.*

fun main(args: Array<String>) {
    val log = DumbLogger()
    log.debug(Supplier<String, Exception> { ioException() })
    log.info<Exception> {
        Thread.sleep(0)
        "This lambda can throw InterruptedException."
    }
}

private fun ioException(): String {
    return "This method declares IOException."
}
```

Hopefully libraries like [kotlin-logging](https://github.com/MicroUtils/kotlin-logging) will find a solution (the `KLogger `interface already extends `org.slf4j.Logger` to add lambda/function overloads of the trace/debug/info/warn/error methods, so adding similar methods to `org.slf4j.Logger` might cause trouble.) 

Any thoughts on this proposal @oshai?

oshai:
personally, I prefer the suggestion of @ceki. declaring the methods as throwing exception is not coherent with to original methods and I haven't seen such signature in other java libs.

BalRoger:
Let me address your reasons one at a time. First, the exception-safe version of these methods is MUCH more coherent with the original usage in terms of actual impact on the utility of code.

In the original API you would enter something like the following:

``` java
log.debug(foo.getBar());
```

And, assuming the surrounding code was able to handle ```IOException```, everything would be fine.

Now, because ```foo.getBar()``` is a somewhat expensive operation, it would be desirable to convert the logging to the lambda-based format:

``` java
log.debug(() -> foo.getBar());
```

Without the exception-safety generic parameterization, this code will fail to compile, regardless of how ```IOException``` is handled in the surrounding code. I find this to be a significant and debilitating change in the coherence (and utility!) of the logging interface.

Second you argue against a more robust solution exception-safe solution because "the standard java libs don't to it that way"? Well, quite frankly I find the the Java-8-defined functional interfaces to be of extremely limited usefulness for precisely that reason. I find quite often that I have to create an exception-safe alternative to a standard interface in ```java.util.function``` just to make it useful for me. If I can't put code that declares checked exceptions in a lambda, about half of all my opportunities to use lambdas to good advantage (including, most especially, those involving logging) won't work.

If the exception-unfriendly solution is adopted, I will continue to use my homegrown logging wrapper that actually works and ignore the half-broken "standard" solution.

xenoterracide:
since the only point of the exception is probably an auto-catch->log instead of making it a generic just make it `get() throws Throwable;` I haven't needed that for logging... but have needed it for `Try` wrappers.

BalRoger:
If you use the generic exception parameter you can convert the logging from current style to lambda-style without having to change the surrounding code.

If you use ```Throwable``` instead of the generic, you have to change the surrounding code to handle ```Throwable```. That is, the following will not suffice:

Given this original code:

``` java
try {
    log.debug(foo.getBar());
} catch (IOException ioe) {
    // whatever ...
}
```

Then changing to lambda-style  will force you to change the surrounding code to handle _**any**_ ```Throwable```, not just ```IOException```. Thus, the minimal change to support lambda-style is now:

``` java
try {
    log.debug(() -> foo.getBar());
} catch (IOException ioe) {
    // whatever ...
} catch (Throwable t) {
    throw new RuntimeException(t);
}
```

No additional changes to the original code are needed if the exception is handled via generics. Well, aside from the current-style to lambda-style change itself.

xenoterracide:
I was thinking that you'd have

     log.debug( ThrowableSupplier supplier ) {
          try {
              log.debug( supplier.get() ) {
          }
         catch ( Throwable e )  {
              log.debug( "debug supplier threw exception {}", e ) 
         }

thus your caller doesn't need to handle the exception thrown by the supplier

BalRoger:
I see your point, but as a programmer I'm not sure I want my logging system intercepting my exceptions. For example, given your code what would happen if debug level is disabled? The exception would be lost!

I personally would rather have the exception behavior with lambdas be as close to the original behavior as I could get.

xenoterracide:
I updated my sample at little, but if you had debug level disabled the supplier would never be called, thus the exception couldn't happen. The point of the lambda's in the first place is to lazily evaluate the supplier, thus no log level, no evaluation. I'm ok with treating a logging exception as internal to the logger... as it's not really part of my application code, but that's a personal preference, otherwise you want want to instead pass 2 functions, a logging `Supplier` and an exception `Consumer`

xenoterracide:
That said you might be better off not having the Exception on the supplier and instead using something like Try in javaslang I've also been writing something similar.

BalRoger:
The supplier would be called regardless of log level in your code as written. Let me tweak it a bit:

``` java
void debug( ThrowableSupplier supplier ) {
     if( !log.isDebugEnabled() ) { return; }
     try {
          log.debug( supplier.get() ) {
     }
     catch ( Throwable e )  {
          log.debug( "debug supplier threw exception {}", e ) 
     }
}
```

Okay, now under _this_ code you have an even better point. But I have still lost control of the exception. If debug level _IS_ enabled the logging _system_ will now log the exception at the debug level. So _NOW_ if my log analysis only looks for exceptions at the warn level or above, I'm out of luck. I'll miss this one.

Furthermore, and this is a minor nitpick, easily fixed, your exception logging statement, as written, does not print out the stack trace, only the exception message. It's a small point but it drives home the fact that my program's exception handling would now be partially usurped by the logging system. This does not seem to me to be an area of responsibility that the logging system should take on.

xenoterracide:
>Okay, now under this code you have an even better point.

fundamentally it's the same code afaik, I presume the compiler would be able to optimize that to the same thing.

> Furthermore, and this is a minor nitpick, easily fixed, your exception logging statement, as written, does not print out the stack trace, only the exception message. 

mm, our system would print out the stack trace in that case if you pass the exception object, so does spring boot.

> so NOW if my log analysis only looks for exceptions at the warn level or above, I'm out of luck.

maybe you shouldn't do that... maybe you should analyze all exceptions, and simply disable logging at certain levels... just a thought, I don't have your use case

> This does not seem to me to be an area of responsibility that the logging system should take on.

sure, though I would argue that perhaps it shouldn't handle checked exceptions at all, [javaslang](http://www.javaslang.io/javaslang-docs/#_try).

alternatively as already mentioned

```
void debug( ThrowableSupplier supplier ) {
     if( !log.isDebugEnabled() ) { return; }
     this.debug( supplier, e -> log.debug(  "debug supplier threw exception {}", e );
}

void debug( ThrowableSupplier supplier, Consumer<Throwable> consume ) {
     if( !log.isDebugEnabled() ) { return; }
     try {
          log.debug( supplier.get() ) {
     }
     catch ( Throwable e )  {
          log.debug( "debug supplier threw exception {}", e ) 
     }
}
```

now you could write

      log.debug(() -> foo.getBar(), e -> throw new RuntimeException(e));

BalRoger:
I think you want to change your 3rd from last line in your 2nd debug method to:

``` java
consumer.accept(e)
```

Yeah, this works. If the consensus is to go this way I will grudgingly accept it. I still think the generic exception parameter is a more elegant solution, and would result in general in more minimal code changes to get roughly the same behavior with lambdas than does this solution.

seanf:
Once you introduce a `Supplier` function which could throw an exception, even a `RuntimeException`, the problem of what to do with that exception does become significant. 

Personally, I like @BalRoger's "generic throws" idea (Kotlin interop notwithstanding), but propagating the logging exception isn't _always_ what you want:

1. Sometimes you just want to make absolutely sure your logging can't break your application, and the last thing you want to do is propagate such an exception.
    I know I hate it when my logging (perhaps added to help me debug a problem) actually causes another problem of its own.

2. Other times you do want to treat _every_ exception as significant (after all, it's probably a bug in your code, even if it is logging code) and abort processing immediately by propagating the exception (perhaps directly, or perhaps wrapped in a `LoggingRuntimeException`).

Perhaps it would be justified to add a parameter to `LoggerFactory.getLogger()`, so that the user can choose a strategy for handling such logging exceptions (or pass in a `ThrowableConsumer`). I wouldn't really want to be passing in an exception handler to every debug logging call.

ctubbsii:
While I like the approach @BalRoger  and @seanf outline, to pass through checked exceptions, I think it might be a bit overkill. I don't think it would be unreasonable to provide a way to use lambdas, constrained to runtime exceptions only. This is consistent with the limitations in the use of lambdas elsewhere in Java 8, such as in streams, and I think it's an acceptable trade-off to have the feature.

kashike:
Any chance this could become a thing soon?

yardus:
So this PR basically got stuck on exception handling?

@seanf  - I see @BalRoger proposal in line with what is available now - when you use a method with checked exception to compute a value to log you must take care of the appropriate try/catch. If the method throws an unchecked exception it may break your application using the functional approach as well as the old, imperative one. There really is no difference.

AFAICS, @BalRoger proposal cleanly resolves the every day logging API usage, with the exception of cumbersome calls from Kotlin. There does not seem to be any other critical deficiencies in this PR which should keep it on hold, right? 

Anyway, it would be nice to get this through, given that Java 8 was EOL-ed this March.

rujiel:
I'm not sure why but lambdas aren't available as params in 2.0.0-alpha... are they supposed to be?

sschuberth:
> A new slf4j-api module is out of the question.

Well, there's [slf4j-api 2.0.x](https://www.slf4j.org/faq.html#changesInVersion200) by now... never say never again 😉 However, it indeed [seem like lazy evaluation of messages did not make it in there](https://www.slf4j.org/apidocs/org/slf4j/Logger.html).

ceki:
See [LoggingEventBuilder](https://www.slf4j.org/apidocs/org/slf4j/spi/LoggingEventBuilder.html) and the [fluent API](https://www.slf4j.org/manual.html#fluent).

sschuberth:
> @sschuberth See [LoggingEventBuilder](https://www.slf4j.org/apidocs/org/slf4j/spi/LoggingEventBuilder.html) and the [fluent API](https://www.slf4j.org/manual.html#fluent).

Ah, interesting approach, thanks for pointing it out @ceki!

PR #84, opened 2014-10-10
MDCCloseable instead of Closeable to avoid throwing of IOException

twillouer:
Avoid throwing IOException in putCloseable...

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -71,7 +71,7 @@ public class MDC {
   /**
    * An adapter to remove the key when done.
    */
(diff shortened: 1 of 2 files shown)


denisa:
Please consider merging this. Having to code for IOException nullifies the advantage of putCloseble()…

mattbishop:
@denisa where is the IOException thrown? I'm not fond of the IllegalArgumentException (NPEs are fine for null values) but I am not seeing the IOE.

ceki:
Makes sense to me as the ```MDCClose.close``` method will never throw an exception.

twillouer:
Yes, sorry @ceki, bad design in the first PR.

In this PR, MDCCloseable.close never send exception.

PR #88, opened 2014-11-01
Updating builds

hazendaz:


slf4j-jcl/pom.xml
```diff
@@ -21,33 +22,11 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
-
     <dependency>
       <groupId>commons-logging</groupId>
       <artifactId>commons-logging</artifactId>
-      <version>1.1.1</version>
+      <version>1.2</version>
```
mattbishop: Not sure about this update. It appears more more than the version # needs to be performed. See #89 for discussion.

slf4j-ext/src/test/java/org/slf4j/dummyExt/XLoggerTest.java
```diff
@@ -107,6 +107,9 @@ public void testThrowing() {
     assertEquals(t, logger.throwing(t));
     assertEquals(t, logger.throwing(XLogger.Level.DEBUG,t));
     assertEquals(2, listAppender.list.size());
+    if (listAppender.list.get(0).getThrowableStrRep() == null) {
```
mattbishop: Why this change?

hazendaz: Its been some time since i did that but i believe it protects against possible null pointer that eclipse noted.

--- Original Message ---

slf4j-jcl/pom.xml
hazendaz: Not sure it compiled ok and test cases workef.  This one is for writing to 1.2 rather than jcl to slf4j.  I have used commons logging 1.2 without issues elsewhere in projects using it but not slf4j without issues.  I assume since unit tests all pass it is good.  I can remove that piece if you want me to.

--- Original Message ---

hazendaz:
@mattbishop I resync'd to latest.  A lot of things in this went away in recent slf4j updates with a couple being questions you had originally.  I do still have this updating the targetted commons-logging of 1.2 which I still think is ok and does not apply to other questions because in those cases it is the other direction (ie commons logging to slf4j).  If you still think that needs backed out I will do that ASAP.  Otherwise, this is a clean improvement to the build.  However, there seems to be some build issues no not caused by this but other changes.  It gets through the main stuff then fails in integration tests due to changes to manifest.mf locatino which wasn't part of this.  Obviously I'd prefer to see a clean build here but just wanted to touch base.  Thanks.

mattbishop:
Travis is failing. Also @ceki your thoughts on this change?

ceki:
This PR is quite extensive. I would prefer a smaller more focused request.

hazendaz:
If I spend the time to make smaller PR requests can I get some commitment that my PR will go somewhere rather than sitting for months?  There have now been a number of times changes I've made or suggested ultimately were made outside of PR requests from me and I don't really feel like wasting time here.  I could have even broken this PR into multiple commits if that made it easier to follow.  This was all about cleaning up the maven scripts to start with.

--- Original Message ---

ceki:
Smaller PRs will be responded to promptly.

hazendaz:
Thank you.  I'll get things started again soon.

--- Original Message ---

PR #97, opened 2015-01-16
Pass arguments to JUL as record parameters

bnorm:
While trying to migrate a project from JUL to SLF4J/Logback, I noticed that argument parameters are not being passed through the SLF4J JDK14 adapter.  Since these arguments are being passed through to Logback -- and I believe through the Log4j adapter to Log4j -- it seems natural that they should be passed through the adapter to JUL as well.

First pull request.  Let me know if there are any special procedures I need to follow.

### DIFF
--- a/slf4j-jdk14/src/main/java/org/slf4j/impl/JDK14LoggerAdapter.java
+++ b/slf4j-jdk14/src/main/java/org/slf4j/impl/JDK14LoggerAdapter.java
@@ -73,7 +73,7 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
      */
     public void trace(String msg) {
         if (logger.isLoggable(Level.FINEST)) {
-            log(SELF, Level.FINEST, msg, null);
+            log(SELF, Level.FINEST, msg, null, null);
         }
     }
 
@@ -94,7 +94,7 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
     public void trace(String format, Object arg) {
         if (logger.isLoggable(Level.FINEST)) {
             FormattingTuple ft = MessageFormatter.format(format, arg);
-            log(SELF, Level.FINEST, ft.getMessage(), ft.getThrowable());
+            log(SELF, Level.FINEST, ft.getMessage(), ft.getArgArray(), ft.getThrowable());
         }
     }
 
@@ -117,7 +117,7 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
     public void trace(String format, Object arg1, Object arg2) {
         if (logger.isLoggable(Level.FINEST)) {
             FormattingTuple ft = MessageFormatter.format(format, arg1, arg2);
-            log(SELF, Level.FINEST, ft.getMessage(), ft.getThrowable());
+            log(SELF, Level.FINEST, ft.getMessage(), ft.getArgArray(), ft.getThrowable());
         }
     }
 
@@ -138,7 +138,7 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
     public void trace(String format, Object... argArray) {
         if (logger.isLoggable(Level.FINEST)) {
             FormattingTuple ft = MessageFormatter.arrayFormat(format, argArray);
-            log(SELF, Level.FINEST, ft.getMessage(), ft.getThrowable());
+            log(SELF, Level.FINEST, ft.getMessage(), ft.getArgArray(), ft.getThrowable());
         }
     }
 
@@ -152,7 +152,7 @@ public final class JDK14LoggerAdapter extends MarkerIgnoringBase implements Loca
      */
     public void trace(String msg, Throwable t) {
         if (logger.isLoggable(Level.FINEST)) {
-            log(SELF, Level.FINEST, msg, t);
+            log(SELF, Level.FINEST, msg, null, t);
         }
     }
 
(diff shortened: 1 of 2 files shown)


bnorm:
I do hope I am not over stepping my bounds, but it looks like the unit tests for the JDK14 adapter are lacking.  There is decent coverage of methods but there is almost no validation on some of the unit tests.  I see two actions that should be taken.

1. Add to the existing unit tests and validate my changes are working as expected.
2. Expand the existing unit tests to provide better validation.

I'm definitely doing number 1 but my fear is that number 2 is beyond the scope of this pull request.  If you agree, I will finish up number 1 and submit it with a potential release note and make 2 a separate pull request.  Or, if you don't care, I'll just do both in this pull request.

Either way, I plan to continue with test development and split (or not) the submission as you wish.

bnorm:
After moving forward with numbers 1 and 2 I have come up with the following unit test located here  https://github.com/bnorm/slf4j/commit/5aa542b4b9e58db262e9c51f82e00ef98452c698.  I left it as a separate branch in case you wanted me to submit a separate pull request.

Code coverage numbers:

Element | Class, % | Method, % | Line, %
---------- | ----------- | -------------- | ---------
JDK14LoggerAdapter | 100% (1/1) | 100% (35/35) | 100% (141/141)
JDK14LoggerFactory | 100% (1/1) | 100% (2/2) | 100% (13/13)
StaticLoggerBinder | 100% (1/1) | 100% (5/5) | 100% (9/9)
StaticLoggerBinder | 100% (1/1) | 100% (5/5) | 100% (9/9)
StaticMarkerBinder | 100% (1/1) | 75% (3/4) | 83% (5/6)

bnorm:
```html
    <p>Changed JDK14 adapter to pass log event arguments through as parameters
    of the JDK14 <code>LogRecord</code></p>
```

bnorm:
Pushed my update to the existing unit tests and also a suggested update to the release notes as requested.

Let me know if there is anything else you would like me to do.

mattbishop:
I'd like to merge this in but it needs a rebase. @bnorm would you mind doing so?

Also did you make a PR for logback to include the release note change?

bnorm:
I've rebased the pull request branch.  I also have a secondary branch with an new unit tests for the JDK14 adapter that you might be interested in.  Please take a look and I will merge it into this pull request if you want me to.  (bnorm/slf4j@fa596899f97547ee333b2c1e032391c754ee2e72)  There are a couple comments about the new tests above as well regarding code coverage.

The release note change, as instructed by the contributing document, is a few comments up.  It said not to include it as part of the pull request commits but have it as a comment.

mattbishop:
It would be good to bring in the JDK14 test, but maybe do a separate PR so we can get this change in.

ceki:
@bnorm writes above:
> While trying to migrate a project from JUL to SLF4J/Logback, I noticed that argument parameters are
> not being passed through the SLF4J JDK14 adapter. Since these arguments are being passed 
> through to Logback -- and I believe through the Log4j adapter to Log4j -- it seems natural that they 
> should be passed through the adapter to JUL as well.

Unfortunately, this reasoning in my opinion, although sensible, is also wrong . Let me explain.

SLF4J supports one and only text formatting scheme which is based on accolades '{}'. There have been repeated request by users to make the formatting scheme configurable, in particular support for java.text.MessageFormat style formatting. This is not currently possible because any additional formatting scheme will break some existing logger calls because the log message passed to the logger is just a String and consequently not configurable.

More concretely, assume project P consists of two external components called A and B. Component A relies on "{}" SLF4J-style formatting while component B relies on java.text.MessageFormat style formatting. If the user configures SLF4J to use java.text.MessageFormat style formatting, then log calls in B will render nicely but calls emanating from A will render as gibberish. Similarly, if the user  configures SLF4J to use '{}' SLF4J-style formatting, then logger calls from A will render nicely but calls from component B will render as gibberish, an undesirable state of affairs to say the least.

Here is a typical parameterized logger call in A
```logger.debug("Hello {}", userName)```   (1)

Here is a typical parameterized logger call in B
```logger.debug("Dear {1}", userName)```   (2)

As I tried to explain above, it is not possible to make both statements (1) and (2) work correctly. The current org.slf4j.Logger methods implicitly presume that only one formatting scheme and there is no way to _efficiently_ let SLF4J know that another formatting scheme should be applied.

Coming back to this pull request, passing arguments parameters to JUL will work nicely for users using JUL-style formatting and slf4j-jdk14 but will fail when the end-user switches to some other logging implementation such as log4j or logback. 

Given the above, I am obliged to revert the changes in this commit. I am very sorry for giving this PR false hope.

ceki:
I'd be happy to discuss this further on the slf4j-dev mailing lists.

bnorm:
I actually completely agree with all of your statements on what style of formatting SLF4J should support.  While making the option configurable would be great for small projects on which nothing depends, it would make it a nightmare for developers trying to combine libraries that use different SLF4J formatting styles.  We are in agreement on that point.

However, this pull request, as far as I can tell, would not effect the message formatting behavior of SLF4J at all.  This change would only allow JUL Handlers to access the parameters passed log method.

I'll try and illustrate this.  Take the following call: `logger.debug("Hello {}", userName)`.  If a developer has the logging implementation set to Logback, they can access the `userName` parameter in an Appender via `event.getArgumentArray()`.  However, if the developer is using slf4j-jdk14, the call `record.getParameters()` in a Handler will return `null` and the parameter `userName` will not be accessible.  For the JUL Handler, the formatting has already been done by slf4j-jdk14, with this change it would now just be receiving parameter information as well.

If you still feel the need to revert this pull request, please do so.  I just wanted to take a moment and make sure it was clear my intention was not to change the formatting behavior in any way.

PR #99, opened 2015-01-27
Fixed BasicMDCAdapter leaks MDC information to non-child threads

lukecwik:
BasicMDCAdapter leaks MDC information to non-child threads
BasicMDCAdapter uses a ConcurrentHashMap which is passed by reference from the parent to the child. All child threads then share the same map reference and any modifications by one child is visible to the parent thread and also all other child threads.

Swapped to creating a copy of the parents map instead of storing reference. Also, since this is expected to be only accessed from within the same thread, swapped to use a hash map instead of a concurrent hash map.

Signed-off-by: Luke Cwik <lcwik@google.com>


slf4j-api/src/main/java/org/slf4j/helpers/BasicMDCAdapter.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2004-2011 QOS.ch
+ * Copyright (c) 2004-2011 QOS.ch, Copyright (C) 2015 Google Inc.
```
mattbishop: Google?

lukecwik: Since I work for Google, all our patches go by an opensource team and they requested that I add that.

ceki: This the first time I encounter this issue. I don't fully understand the implications, just enough to be uncomfortable.

lukecwik: Everyone who writes code, owns the copyright and is the author. When someone gives you a patch, they are just *licensing* it to the project maintainer, not transferring ownership.

In general many opensource projects have a contributor licensing agreement (CLA) which details the license under which the terms that the intellectual property has been contributed. It doesn't seem as though SLF4J has such a CLA which stipulates the terms.

More details are here:
http://en.wikipedia.org/wiki/Contributor_License_Agreement

Here are some example CLA agreements for projects which receive many individual contributions:
http://www.apache.org/licenses/icla.txt
https://www.python.org/psf/contrib/contrib-form/
http://www.eclipse.org/legal/CLA.php
https://docs.google.com/a/google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ

Many projects have a file which details the terms for contributing to them, for example:
https://github.com/apache/ant/blob/master/CONTRIBUTING.md which stipulates that a CLA must by signed unless the patch is trivial.

lukecwik: Any updates?

stephenh:
Looks like this is the new bug:

http://jira.qos.ch/browse/SLF4J-316

I think this commit looks great. I'm seeing very weird behavior from our MDC output (values getting mixed matched which don't make sense unless there is really one map being shard everywhere).

I was poking around at InheritableThreadLocal, saw childValue, and googling for "slf4j mdc childValue", found this pull request.

+1 for merging it and releasing.

slf4j-api/src/main/java/org/slf4j/helpers/BasicMDCAdapter.java
mattbishop: Luke it looks like @ceki won't take this PR with the Google copyright.

lukecwik:
Updated to remove the Google copyright. Irrespective of the copyright being there or not, everyone who writes code, owns the copyright and is the author. When someone gives you a patch, they are just licensing it to the project maintainer, not transferring ownership.

slf4j-api/src/main/java/org/slf4j/helpers/BasicMDCAdapter.java
```diff
@@ -86,10 +95,11 @@ public void put(String key, String val) {
   /**
    * Get the context identified by the <code>key</code> parameter.
    */
+  @Override
   public String get(String key) {
-    Map<String, String> Map = (Map<String, String>) inheritableThreadLocal.get();
+    Map<String, String> Map = inheritableThreadLocal.get();
```
mattbishop: I know this was pre-existing, but the variable is capitalized Map, but should be just map.

slf4j-api/src/main/java/org/slf4j/helpers/BasicMDCAdapter.java
```diff
@@ -71,13 +79,14 @@ static boolean isJDK14() {
...
+  @Override
   public void put(String key, String val) {
     if (key == null) {
       throw new IllegalArgumentException("key cannot be null");
     }
-    Map<String, String> map = (Map<String, String>) inheritableThreadLocal.get();
+    Map<String, String> map = inheritableThreadLocal.get();
     if (map == null) {
-      map = Collections.<String, String>synchronizedMap(new HashMap<String, String>());
+      map = new HashMap<String, String>();
```
mattbishop: The map is no longer synchronized, why is now unnecessary to do so?

lukecwik: Since every new thread inherits a copy of its parents map when the child thread is constructed.
Also, childValue method is called from within the parents thread during child thread construction.

slf4j-api/src/main/java/org/slf4j/helpers/BasicMDCAdapter.java
lukecwik: Done

stephenh:
FWIW I've been running this patch in production for ~a week and our MDC output looks much better.

mattbishop:
Looking through this PR it is clear that the map managed by the ThreadLocal never escapes the class so the synchronized map is not required, just as @lukecwik stated.

I'd like to merge it now but it has conflicts with master that need resolving.

mattbishop:
Thanks Luke. I want to merge this in now. Any objections?

mattbishop:
This bug is in Jira now: http://jira.qos.ch/browse/SLF4J-316

ceki:
Given its size, this contribution requires a signed [ICLA](http://logback.qos.ch/cla.txt). 


@lukecwik If you think that you can't sign the ICLA, please let us know so that we can find another way for fixing http://jira.qos.ch/browse/SLF4J-316

lukecwik:
Taking a look at ICLA, will give response soon as to whether I can sign it.

lukecwik:
I have signed the ICLA and e-mailed a copy to ceki.
Waiting on ceki to confirm that he has received it and everything is in order.

ceki:
The CLA looks good. Thank you Lukasz.

mattbishop:
...and it is out of date again. Please rebase one last time, ftw.

mattbishop:
@lukecwik Can you please rebase this PR and I will merge it in?

lukecwik:
Done, please merge PR when able.

lukecwik:
Ping?

ceki:
Ill look at this on my return on the 25th of July.

SvenBunge:
As I see the same fix needs to be applied in logback as well: 
See https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/util/LogbackMDCAdapter.java

alf:
Ping.

lukecwik:
Noted that the LogbackMDCAdapter leaks values to non-child threads but Logback is in a different project and will require a separate pull request.

The LogbackMDCAdapter also copies the map within the same thread for every read/write operation combo.

Also ping for merging this in.

PR #100, opened 2015-02-16
Caching logger references in bridges can break underlying contextual logging creation

aartigao:
If we have a shared library in our Application Server with different J2EE applications running in the same JVM, contextual logging creation is broken with cached references in SLF4J Bridges. 

A classical example is JUL in Spring Framework and ContextJNDISelector. When an application boots, the first class that loads a Logger through JUL, will expose this logger to all applications running in the server (JdbcTemplate is a great example).

By forcing always the SLF4J  Logger reference object, we delegate the caching to the underlying SLF4J implementation, using this Logger as a key for Log map.

### DIFF
--- a/jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
+++ b/jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
@@ -56,7 +56,7 @@ public class SLF4JLogFactory extends LogFactory {
 
   /**
    * The {@link org.apache.commons.logging.Log}instances that have already been
-   * created, keyed by logger name.
+   * created, keyed by Logger instance.
    */
   ConcurrentMap<Logger, Log> loggerMap;
 
--- a/jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
+++ b/jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
@@ -17,11 +17,14 @@
 package org.apache.commons.logging.impl;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.List;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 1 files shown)


jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
```diff
@@ -56,15 +56,15 @@
 
   /**
    * The {@link org.apache.commons.logging.Log}instances that have already been
-   * created, keyed by logger name.
+   * created, keyed by Logger instance.
    */
-  ConcurrentMap<String, Log> loggerMap;
+  ConcurrentMap<Logger, Log> loggerMap;
```
mattbishop: What happens when the webapp the loaded Logger is unloaded? loggerMap will still retain a reference, no?

Maybe use WeakHashMap with a ReadWriteLock to control access.

aartigao:
That's a good point! Maybe we can control access with Collections.synchronizedMap(WeakHashMap). WDYT?

jcl-over-slf4j/src/main/java/org/apache/commons/logging/impl/SLF4JLogFactory.java
aartigao: That's a good point! Maybe we can control access with Collections.synchronizedMap(WeakHashMap). WDYT?

aartigao: Anyway, I see, we need a ReadWriteLock for loggerMap.putIfAbsent :)

ceki:
Can you describe how the WeakHaspMap help solve the problem?

mattbishop:
@ceki I see your point. The weak hash map would drop the Logger in more cases beyond app unload.

aartigao:
@ceki @mattbishop Should I push a version without intermediate cache? Returning a new wrapper SLF4JLocationAwareLog/SLF4JLog reference always?

ceki:
It is not clear to me whether returning a new wrapper would be acceptable in all circumstances. For example, in case of non-static logger references, the log4j paradigm (by ricochet slf4j) always returns the same logger instance for the same logger name. I would not want to change a cornerstone behavior without further investigation. I

I would be OK with a system property which when set would avoid caching of wrapper instances. The default behavior would remain unchanged.

Again, it might be OK to return a new wrapper on each getLogger call, we just need to think about it more.

aartigao:
I understand the problematic. If there's something that I can help, please, let me know. Thanks for your attention :smiley:

PR #101, opened 2015-03-12
Update AppenderSkeleton to include setThreshold() method mute implementation

dimitrisli:
Adding and muting the setThreshold method call without which the Log4j bridge is broken for underlying dependencies that are manipulating Log4j Appenders directly


dimitrisli:
This is following the implementation footsteps of org.apache.log4j.Logger#addAppender() introduced in v.1.7.3. If this method is missing then a java.lang.NoSuchMethodError is getting thrown

dimitrisli:
I've created ticket: http://bugzilla.slf4j.org/show_bug.cgi?id=329 for this issue to link to this pull request

mattbishop:
Pretty standard.

PR #103, opened 2015-04-01
Add xml version / encoding utf-8 to poms and format project tag

hazendaz:
This change is only to ensure xml version and utf-8 are tagged at start of poms to comply with proper structure.  The project tag is formatted to ensure consistancy across the library.  This change has no impact to the build process itself.  Its purpose is for consistancy.


ceki:
@hazendaz @mattbishop The change has no impact because default encoding in XML documents UTF-8. It that were not the case, and if one of the pom files was relying on a different (default) encoding, the change would cause problems difficult to debug.

In general, I am in favor of housekeeping changes but they still need to be considered with care.

PR #105, opened 2015-04-03
Update comments to point to JIRA equivalents of old Bugzilla bugs

jmthibault79:


mattbishop:
rebase?

jmthibault79:
Rebased.

mattbishop:
Build is failing on Travis.

jmthibault79:
Fixed: a test was dependent on its own line numbers, so it needed an update after I added a comment line.

mattbishop:
Actually why are we keeping bug references in the code at all?

jmthibault79:
I was able to solve a problem I had with using SLF4J thanks to one of these bug comments - once I was able to map its Bugzilla URL to the new JIRA ticket.

That's my motivation, anyway.

ceki:
@mattbishop Comments in the code mention a bugzilla/JIRA issues for further details on the context of the problem. The line number issue is quite different. The test in question verifies line number extraction and hence depends on the caller's line number. It's quite fragile but the fix is also quite easy.

@jmthibault79  As bugzilla.slf4j.org is no longer available, this patch was highly welcome. Many thanks.

PR #107, opened 2015-04-08
Apply Standard Maven Usage

hazendaz:
Add additional UTF-8 settings, use project.basedir, and move modelVersion to proper location.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,7 @@
 
     <resources>
       <resource>
-        <directory>src/main/resources</directory>
+        <directory>${project.basedir}/src/main/resources</directory>
         <filtering>true</filtering>
       </resource>
     </resources>
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,9 @@
     <junit.version>4.10</junit.version>
     <maven-site-plugin.version>3.3</maven-site-plugin.version>
     <javadoc.plugin.version>2.10.2</javadoc.plugin.version>
+
+    <signature.artifact>java15</signature.artifact>
+    <signature.version>1.0</signature.version>
   </properties>
 
   <developers>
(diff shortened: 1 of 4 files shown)


slf4j-site/pom.xml
```diff
@@ -20,7 +20,7 @@
   <build>
     <resources>
       <resource>
-        <directory>src/site/pages</directory>
+        <directory>${project.basedir}/src/site/pages</directory>
         <!--We're saving filtered html docs in a temporary folder-->
         <!--and telling the site plug in to get the docs there.-->
         <targetPath>../../../target/site</targetPath>
```
mattbishop: Why isn't this ${project.basedir}/target/site ?

hazendaz: In that location it would look in the wrong place.  The base directory is the module it is in.  In this case that info is outside of that module.  I can take at making that part of the build more maven friendly in another PR.

--- Original Message ---

hazendaz: Actually that might work there with a site adjustment.  I will check tonight.

--- Original Message ---

hazendaz:
I'll fix the merge conflicts shortly if anyone is watching at the moment.

hazendaz:
ok - merge conflict was me addressing request to explicitely state jdk and @ceki making it sometime after this morning.  I reapplied his changes and just kept my update of the compiler plugin.  Will be pushing back up in a moment.

hazendaz:
This pull has items @ceki was after for animal sniffer.

hazendaz:
Any status on this?

pom.xml
```diff
@@ -122,16 +126,47 @@
...
         <filtering>true</filtering>
       </resource>
     </resources>
 
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>animal-sniffer-maven-plugin</artifactId>
+          <version>1.14</version>
```
mattbishop: Most of the other plugins have their version # propertized.

slf4j-api/pom.xml
```diff
@@ -21,6 +21,19 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>animal-sniffer-maven-plugin</artifactId>
+        <configuration>
+          <ignores>
+            <ignore>org.slf4j.impl.StaticMDCBinder</ignore>
```
mattbishop: Why are these ignored? Perhaps add a comment for each.

hazendaz: Comment is on the commit but I see your point that it would be useful in the file.  I'll add that to this as soon as I get a chance.

--- Original Message ---

pom.xml
hazendaz: I add the version to properties as soon as I can.

--- Original Message ---

hazendaz:
@mattbishop 1 for 2.  I added comments as to why animal sniffer has to ignore the interfaces or classes specifically marked that they shouldn't be in users code.  It wouldn't work with those as it couldn't find a concrete class in those specific modules.  I left the version in the location it is because the vaste majority of plugins in this library do not comply to using properties and I felt that maybe that warrants a separate PR to clean them all up in one go.

PR #108, opened 2015-04-08
SLF4J-312 Using AbstractStringBuilder#append(str,start,end)

dimitrisli:
SLF4J-312 Eliminating one stack frame by using AbstractStringBuilder#append(str,start,end). A bit more readable compared to inlined String.substring(start,end)


dimitrisli:
Not amazing value added, I am currently looking into open issues to see how I can start contributing.

mattbishop:
A small change but the code is clearer to me without the substrings. Nice not to do the unnecessary.

PR #110, opened 2015-04-15
SLF4J-218 removed org.apache.log4j.spi.Layout should be without "spi"

panchenko:
[SLF4J-218](http://jira.qos.ch/browse/SLF4J-218) Removed duplicated Layout from "spi", in log4j it is [org.apache.log4j.Layout](https://logging.apache.org/log4j/1.2/apidocs/index.html?org/apache/log4j/Layout.html) which is already present.


panchenko:
The correct file is here [Layout.java](https://github.com/qos-ch/slf4j/blob/master/log4j-over-slf4j/src/main/java/org/apache/log4j/Layout.java)

mattbishop:
Log4J Layout: https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Layout.html

SLF4J has two identical Layout classes. @ceki any reason to keep the *.spi.* version?

ceki:
@mattbishop There is no reason to keep the spi version. I think the original idea was to tuck Layout somewhere hidden but it was later requested and placed in its proper place, i.e. the org.apache.log4j package. However, the copy in org.apache.log4j.spi was not removed.

PR #111, opened 2015-04-18
BasicMDCAdapter: use cached value of isJDK14()

panchenko:


mattbishop:
Do you have any perf metrics to show this is an improvement? Usually the JIT will optimize these kinds of calls at runtime if the test always returns the same answer with the same data.

panchenko:
a) That looks lie an initial not finished idea, as that variable is not used at the moment
b) I don't think JIT can optimize this function call - that expression can't be simplified to a constant

SvenBunge:
Please correct me but I think JDK1.4 isn't supported anymore since 1.7. I've removed the JDK1.4-case completly in my PR #119 
See http://www.slf4j.org/faq.html#requirements

panchenko:
Makes sense.

PR #112, opened 2015-04-18
Use _ to shorten tag names instead of *.

lhunath:
A common method of enabling logging for a certain tag involves issuing
the setprop command via adb shell.  Unfortunately, it is impossible to
set a property on a key whose name contains a \* using this method.  As
such, \* is a bad choice for shortening tag names, since it means we
cannot use this method for setting the log level of a tag.

### DIFF
--- a/slf4j-android/pom.xml
+++ b/slf4j-android/pom.xml
@@ -11,6 +11,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>slf4j-android</artifactId>
+    <version>1.7.13-underscore</version>
     <packaging>jar</packaging>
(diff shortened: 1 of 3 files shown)


mattbishop:
Can you point to some android docs that describe this convention?

slf4j-android/pom.xml
```diff
@@ -20,7 +21,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>${project.version}</version>
```
mattbishop: Can you revert the POM changes? I don't think you meant to include them in the PR.

PR #113, opened 2015-04-22
Update Category.java

mciezadlo:
Added missing method from log4j Category class, which is quite useful and can be easily implemented.
(Found it when using library which called this method so I received NoSuchMethodError)


mattbishop:
This method is in Log4J: https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Category.html#assertLog(boolean,%20java.lang.String)

PR #114, opened 2015-04-23
Lazy load SecurityManager.

TobiX:
This fixes SLF4J-324, since the user can opt out of the
slf4j.detectLoggerNameMismatch if there is a SecurityManager that denies
"createSecurityManager".

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/helpers/Util.java
+++ b/slf4j-api/src/main/java/org/slf4j/helpers/Util.java
@@ -46,7 +46,7 @@ public final class Util {
         }
     }
 
-    private static final ClassContextSecurityManager SECURITY_MANAGER = new ClassContextSecurityManager();
+    private static ClassContextSecurityManager SECURITY_MANAGER = null;
 
     /**
      * Returns the name of the class which called the invoking method.
@@ -54,6 +54,11 @@ public final class Util {
      * @return the name of the class which called the invoking method.
      */
     public static Class<?> getCallingClass() {
+        synchronized (Util.class) {
+            if (SECURITY_MANAGER == null) {
+                SECURITY_MANAGER = new ClassContextSecurityManager();
+            }
+        }
         Class<?>[] trace = SECURITY_MANAGER.getClassContext();
         String thisClassName = Util.class.getName();
 


mattbishop:
I don't like this fix because it masks an actual problem with using ClassContextSecurityManager. Furthermore it doesn't help anyone using the new autodetecting Logger feature in an environment that does not let you create a security manager because they won't know how to fix it. Worse, they won't be able to fix it.

The util should check the security permissions before instantiating the ClassContextSecurityManager. If permission isn't granted, then the util could use the slower Exception-based method.

@adorokhine what do you think?

adorokhine:
We would want to have this functionality work in some form under Websphere
--- hopefully we'll still eventually get support for
anonymously-constructed loggers, and those should work under websphere too.

I'll try your suggestion and/or see if there's another solution. Thanks for
bringing it to my attention.

adorokhine:
Mailed pull request #115 implementing Matt's suggestion to fall back to
exceptions. Also includes a write-up of alternatives considered.

mattbishop:
Just curious, did you ever try out sun.reflect.Reflection.getCallerClass()? I know it's not multi-java aware, but I would be curious to see it as an option for those VMs that have this class.

adorokhine:
@mattbishop I'm assuming this comment is for me...

I've been thinking about how it might be possible to do this. With the new refactoring of PR #115, I think it might be possible to try loading such a class dynamically through reflection, assuming the security settings allow it. If it works, it can be called through the interface like a normal class. If it doesn't work, we can fall back to a different implementation.

There are a few problems however:
1. The resulting compiled .jar would still work on non-sun JVMs if we try to load the class dynamically, but I don't think slf4j would compile from source anymore on these JVMs, because the class file would have to directly import sun.reflect.* classes which wouldn't be resolved.
2. I believe the security manager can be configured to deny loading classes at runtime like this. I'm not sure what the most common settings are, but I fear that if security manager creation is disabled, reflection and classloader stuff might be disabled as well.
3. (most importantly) The sun.reflect.* classes change from release to release in unpredictable ways. I looked at these particular classes before, and there used to be a method which returned the caller at an arbitrary stack depth. Between Java 6 and 7, they no longer needed this functionality, so they changed this method to just immediately throw an UnsupportedOperationException. If we had been using this functionality, slf4j would jave just failed to work against Java 7+ until we updated the code to some new method and issued a new release. I fear this would make the code fragile and tightly coupled to a particular Java version.

Based on the above, I feel like it probably wouldn't be a good idea to try this approach. But what do you think? I can try implementing something like this as an experiment for us to play with if you still think it would be useful.

Thanks for all your help.

Cheers,
Alex.

mattbishop:
This PR is superceded by #115.

TobiX:
Right, closing.

PR #117, opened 2015-05-02
Move MANIFEST build to Parent

hazendaz:
Add two empty manifest files for integration and slf4j-site
Remove all pieces of manifest building on jar that can be in parent

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -177,6 +177,25 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
         <version>2.3.1</version>
+        <configuration>
+          <archive>
+            <manifest>
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+              <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+            </manifest>
+            <manifestEntries>
+              <Bundle-Version>${parsedVersion.osgiVersion}</Bundle-Version>
+              <Bundle-Description>${project.description}</Bundle-Description>
+              <Os-Name>${os.name}</Os-Name>
+              <Os-Arch>${os.arch}</Os-Arch>
+              <Os-Version>${os.version}</Os-Version>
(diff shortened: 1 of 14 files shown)


hazendaz:
The resources add for site page module ensures it builds without failing as it otherwise won't include the manifest.

pom.xml
```diff
@@ -177,6 +177,25 @@
...
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+              <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+            </manifest>
+            <manifestEntries>
+              <Bundle-Version>${parsedVersion.osgiVersion}</Bundle-Version>
+              <Bundle-Description>${project.description}</Bundle-Description>
+              <Os-Name>${os.name}</Os-Name>
+              <Os-Arch>${os.arch}</Os-Arch>
+              <Os-Version>${os.version}</Os-Version>
+              <X-Compile-Source-JDK>${maven.compiler.source}</X-Compile-Source-JDK>
```
mattbishop: Why are X-Compile... added to the manifest?

hazendaz: It tells users what java version library targets.  Much easier than assumptions.  It clearly shows binary compatibility.

--- Original Message ---

ash2k: Is it part of any standard? There is [Bundle-RequiredExecutionEnvironment](http://wiki.osgi.org/wiki/Bundle-RequiredExecutionEnvironment) which is part of OSGi. If you decide to change to it, make sure you use correct string from http://wiki.osgi.org/wiki/Execution_Environment.

hazendaz: Apache does it for their commons modules.  I've seen it used in lots of other places over the years.  I don't know if it is for any specific standard.  It has nothing to do with OSGi.  This from what I understand makes it easy for me as a user of a library to know if it will work against a given JVM.  Without it, the build just tells us the product was built with some version of java.  If that version of java is higher than we use, one questions if it can be used.  This clears up that confusion.  I have no idea if this assists for anything otherwise other than I had read long ago it was good practice.

PR #118, opened 2015-06-02
Improvement to MDC and MDCCloseable: Multiple key-value pairs

kutsal:
This commit adds support for multiple key-value pairs to be specified in
the `putClosable` method of the `MDC` class. It modifies `MDCClosable`
class to keep track of the added keys and supports their removal upon
call to the `close` method in a _try-with-resources_ block:

``` java
try(MDCClosable c = MDC.putCloseable("key1", "value1", "key2", "value2", ...)) {
   ...
}
// Keys we put are now removed
```

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -25,7 +25,10 @@
 package org.slf4j;
 
 import java.io.Closeable;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
 
 import org.slf4j.helpers.NOPMDCAdapter;
 import org.slf4j.helpers.BasicMDCAdapter;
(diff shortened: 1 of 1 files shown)


mattbishop:
I think we are pinned to JDK 1.5 so we can't do try-with-resources. Nice though!

kutsal:
@mattbishop yes, of course, the try-with-resources block is not usable in 1.5. But `Closable` is, and this commit builds on top of a class that already exists in slf4j (`MDCClosable` was introduced in 1.7.8), and I wanted to make that bit a little more usable for those of us on Java 1.7 and above.

madrob:
Glad I checked the existing PRs - I was going to make this same suggestion!

@kutsal - One implementation suggestion... Maybe have two separate methods? One that takes `(key,value)` and one that is full vararg. That way putting a single value in doesn't have the vararg method performance impact.

PR #119, opened 2015-06-02
Use Java5 features and remove legacy code

SvenBunge:
Since SLF4J 1.7.x a JDK of version 5 or higher is required. Due previous versions old-fashioned javacode is still in the API module of slf4j, which I try to replace with Java 5 APIs and language features.

Removes/changes of this PR:
* legacy check for JDK1.4 to choose correct cleanup method for ```InheritableThreadLocal```
* casts of values retrieved from collections 
* ```String#indexOf``` replaced by ```String#contains```
* Use for-each loops where possible

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -126,9 +126,9 @@ public final class LoggerFactory {
     private static boolean messageContainsOrgSlf4jImplStaticLoggerBinder(String msg) {
         if (msg == null)
             return false;
-        if (msg.indexOf("org/slf4j/impl/StaticLoggerBinder") != -1)
+        if (msg.contains("org/slf4j/impl/StaticLoggerBinder"))
             return true;
-        if (msg.indexOf("org.slf4j.impl.StaticLoggerBinder") != -1)
+        if (msg.contains("org.slf4j.impl.StaticLoggerBinder"))
             return true;
         return false;
     }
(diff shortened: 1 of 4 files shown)


mattbishop:
This looks fine to me. It will break a few other PRs but it's worth it to get the JDK14 test out.

slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
```diff
@@ -126,9 +126,9 @@ private final static void performInitialization() {
     private static boolean messageContainsOrgSlf4jImplStaticLoggerBinder(String msg) {
         if (msg == null)
             return false;
-        if (msg.indexOf("org/slf4j/impl/StaticLoggerBinder") != -1)
+        if (msg.contains("org/slf4j/impl/StaticLoggerBinder"))
```
mattbishop: Verified introduced in JDK 1.5.

slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
```diff
@@ -198,8 +198,8 @@ private final static void versionSanityCheck() {
             String requested = StaticLoggerBinder.REQUESTED_API_VERSION;
 
             boolean match = false;
-            for (int i = 0; i < API_COMPATIBILITY_LIST.length; i++) {
-                if (requested.startsWith(API_COMPATIBILITY_LIST[i])) {
+            for (String aAPI_COMPATIBILITY_LIST : API_COMPATIBILITY_LIST) {
```
mattbishop: for-each added in Java 5: http://docs.oracle.com/javase/1.5.0/docs/guide/language/foreach.html

PR #120, opened 2015-07-12
Add lifecycle-mapping ignore for plugin execution

hazendaz:
maven-antrun-plugin added with execution filter set to ignore under
usage in eclipse only.  This plugin's configuration is used to store
Eclipse m2e settings only.  It has no influence on the Maven build
itself.

### DIFF
--- a/slf4j-api/pom.xml
+++ b/slf4j-api/pom.xml
@@ -81,7 +81,47 @@
         </configuration>
       </plugin>
     </plugins>
-
   </build>
 
+  <profiles>
+    <profile>
+      <id>eclipse</id>
+      <activation>
+        <property>
+          <name>m2e.version</name>
+        </property>
+      </activation>
+      <build>
+        <pluginManagement>
+          <plugins>
+            <!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
+            <plugin>
+              <groupId>org.eclipse.m2e</groupId>
+              <artifactId>lifecycle-mapping</artifactId>
+              <version>1.0.0</version>
+              <configuration>
+                <lifecycleMappingMetadata>
+                  <pluginExecutions>
+                    <pluginExecution>
+                      <pluginExecutionFilter>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <versionRange>[1.3,)</versionRange>
+                        <goals>
+                          <goal>run</goal>
+                        </goals>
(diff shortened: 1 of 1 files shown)


redrebel:
this is usefull commit. 
i think.....:)

mattbishop:
Yes and we have no unit tests that spin up an OSGi container to see if the bundles load correctly.

hazendaz:
Confused?  What does OSGI have to do with making Eclipse not show an error?

Per execution filter which is auto generated in eclipse...

hazendaz:
@mattbishop Can I get a status on this?  I don't understand your original comment.  Execution filters are for eclipse usage and have no impact to builds in any way.  It just pevents showing errors and this was for ant.  You had mentioned something about OSGi which I don't see how that matters.  This is just for us eclipse users because m2e is such a mess.  For anyone using anything else to build including maven itself, this doesn't even activate.

mattbishop:
@hazendaz the m2e was a red herring. I fixed this issue by using the manifest templates used elsewhere instead of maven-bundle-plugin.

mattbishop:
I mean to say I fixed the other OSGi issue not this issue!

hazendaz:
@mattbishop Since this was left as still outstanding, can we get it applied?

ceki:
@hazendaz I use Eclipse daily (but not m2eclipse). Can you describe the error in more details. Does it occur only with m2eclipse? If so, what is the nature of the error?

hazendaz:
Eclipse flags it with a lifecycle error so this just disables m2e from seeing.  Lots of plugins have this issue.  Generally its trying to tell user to install m2e plugin but those are not always available or necessary.

hazendaz:
I should further note, if you are using eclipse and have a maven project, you are in fact using m2e.  That is how it works with eclipse.  There use to be the eclipse plugin but if that is still used on any importing to eclipse it messes things up.  Maven team retired that plugin quite a while ago.  All modern versions of eclipse comes with m2e out of the box.  Importing the project is going to use m2e to accomplish it.

Possible reasons you might not see the error.

- If you disabled the checks to flag it as an error
- If you allowed it to be added to the ignores within m2e in eclipse

Simple way to find that out is to simply check the configuration on maven and confirm if it was flagged there.  This fix goes one step further and just makes it happen at the project level so no users ever see the error.

hazendaz:
@ceki Do we want to do anything with this?  Possibly a jira or something like that needed to get this in?  I've got other changes behind it that I just haven't raised.

ceki:
fixed in  https://github.com/qos-ch/slf4j/commit/e0120c1473

PR #122, opened 2015-07-23
Binding Ambiguity check skipped for Android

nitinverma:
As Dex cant by design have duplicate class, there would never be a 'Binding Ambiguity'.

1. LoggerFactory.getLogger is mostly called via MainActivity. Which gives a black screen before the first activity loads.
2. This operation takes 200-600 Inclusive CPU Time as per DDMS.

This fix get Incl CPU Time to sub 10

TestCase: org.slf4j.DetectMultipleBindingsTest
Using 'slf4j.detectMultipleBindings' system property as per @mattbishop's feedback.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -84,6 +84,8 @@ public final class LoggerFactory {
 
     // Support for detecting mismatched logger names.
     static final String DETECT_LOGGER_NAME_MISMATCH_PROPERTY = "slf4j.detectLoggerNameMismatch";
+    static final String JAVA_VENDOR_PROPERTY = "java.vendor.url";
+        
     static boolean DETECT_LOGGER_NAME_MISMATCH = Util.safeGetBooleanSystemProperty(DETECT_LOGGER_NAME_MISMATCH_PROPERTY);
 
     /**
@@ -254,6 +256,11 @@ public final class LoggerFactory {
      *
      */
     private static void reportMultipleBindingAmbiguity(Set<URL> staticLoggerBinderPathSet) {
+        if(isAndroid()) {
+            // skip check under android, see also http://jira.qos.ch/browse/SLF4J-328
+            return;
+        }
+        
         if (isAmbiguousStaticLoggerBinderPathSet(staticLoggerBinderPathSet)) {
             Util.report("Class path contains multiple SLF4J bindings.");
             for (URL path : staticLoggerBinderPathSet) {
@@ -261,6 +268,13 @@ public final class LoggerFactory {
             }
             Util.report("See " + MULTIPLE_BINDINGS_URL + " for an explanation.");
         }
+    }
+
+    private static boolean isAndroid() {
+        String vendor = Util.safeGetSystemProperty(JAVA_VENDOR_PROPERTY);
+        if(vendor == null)
+            return false;
+        return vendor.toLowerCase().contains("android");
     }
 
     private static void reportActualBinding(Set<URL> staticLoggerBinderPathSet) {
--- a/slf4j-site/src/site/pages/news.html
+++ b/slf4j-site/src/site/pages/news.html
@@ -52,7 +52,12 @@
    provided the relevant pull request.
    </p>
 
-   
+   <p>During initialization the binding ambiguity check is skipped
+   under Android in order to improve performance. This change was
+   requested by Nitin Verma in <a
+   href="http://jira.qos.ch/browse/SLF4J-328">SLF4J-328</a> who also
+   provided the relevant patch.
+   </p>
 
    <p>The <code>StaticMarkerBinder</code> and
    <code>StaticMDCBinder</code> classes shipping in various SLF4J


nitinverma:
Please refer: http://jira.qos.ch/browse/SLF4J-328

digitalbuddha:
LGTM.  Would love to see this merged soon. We use SLF4J in the NYTimes android app, we like it because we can also use it in our libraries that do not include android sdk.   During startup we have a [100ms+ delay](https://www.nimbledroid.com/play/com.nytimes.android?p=2YQttV9kW2kGgY#LoggerFactory.findPossibleStaticLoggerBinderPathSet) caused by calls to ```StaticLoggerBinderPathSet```.  Calls to ```Classloader.getResources``` are terribly slow on Android, its a shame that such a great library unnecessarily slows down startup time.

nitinverma:
Thanx for reenforcing the benefits of the patch, Mike ( @digitalbuddha )
rebased after Pull Request after 1.7.13 release.

ceki:
@nitinverma @digitalbuddha If possible I'd like to avoid introducing a new system property. How about performing the check in a new thread?

nitinverma:
Yes, we can try to alter the code to do this check async. Alternatively, we could just detect if it's android and avoid this system property.
( https://github.com/nitinverma/slf4j/commit/4d67b578fdc1cf77a0689f2efb0424f6c8479963 )

Out of the 3 alternatives. I would favor 'system property'
1. As it eliminated this check which is unnecessary on android. Saves valuable battery and startup time. (no additional cost of thread creation/destruction too)
2. Gives the control to developer. no magic.
3. Preserves current behaviour by default. no surprises. 

Please suggest.

ceki:
@nitinverma I am sorry for the dissonant voices. Matt suggesting one approach and me another. 

As for my preferences, I think that if we can quietly do the right thing under all circumstances, than we should aim for that. Only if that is not possible, should we involve the user (via a system property).

Checking for "android" in the "java.vendor.url" system property is simple (=good) but only solves the problem for android. I suspect the problem also affects other environments, this the asynchronous approach is worth a try,

nitinverma:
@ceki Sure, I'll create a new branch on my fork and do changes for async check.

nitinverma:
After tests on Android (MI3W, 4.4.4, QuadCore 2.3GHz, 2GB RAM)

1. Thread creation cost has high variation 13988594, 5745520, 3525156, 1115938, 1364896, 875521, 222448 nanoseconds.
2. Only first call to findPossibleStaticLoggerBinderPathSet has significant cost, subsequent calls are 30-40 times cheaper. Thus averaging calls in a loop doesn't give the right picture.   
3. findPossibleStaticLoggerBinderPathSet call has a lower variation. Found between 4505937-1269895 nanoseconds.

Relevant Code:
https://github.com/nvexperiments/Slf4jBindingAmbiguityCheck/blob/develop/apps/slf4j-app/src/sync/java/edu/nitin/slf4j/app/Flavor.java
https://github.com/nvexperiments/Slf4jBindingAmbiguityCheck/blob/develop/apps/slf4j-app/src/async/java/edu/nitin/slf4j/app/Flavor.java

You can build the apks from https://github.com/nvexperiments/Slf4jBindingAmbiguityCheck or I can build and checkin the same.

ceki:
@nitinverma  BTW, how can you be sure that the multiple binding detection is superfluous on android?

digitalbuddha:
Dex does not allow multiple of same class on class path. It will bomb during compilation.

ceki:
@digitalbuddha  OK. And you can't have multiple dex files?

digitalbuddha:
Even with multidex they will be merged back together. We have the duplicate class problem come up if we include a dependency within a library project and then again in main project. That  is with multidex enabled. Let me know if there's something you'd like verified on our code base. We have slf4j in 3 android archives,  2 java archives and a main project which all builds out to a single apk.

ceki:
@digitalbuddha My apologies I am not familiar with android development. I would only like to ascertain that SLF4J's multiple binding detection is superfluous on android.

digitalbuddha:
Mine as well... I'm not very familiar with slf4j multiple bindings. We just
noticed the slowdown in our app and found this related issue/pr. Let me
dive deeper tonight and try to force multiple bindings to see what happens.
Again thank you for a timeless  library.

nitinverma:
@ceki @digitalbuddha 
Android even while generating APK in multidex mode does the dup-class check at compile time.

Experiment (with multidex generation):
Added org.slf4j.LoggerFactory to my source with slf4j as dependency.
https://github.com/nvexperiments/Slf4jBindingAmbiguityCheck/commit/5e6cc8e1fb311fcd58aaf33c16baed19ecc890f8
===
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':apps:slf4j-app:packageAllAsyncDebugClassesForMultiDex'.
> java.util.zip.ZipException: duplicate entry: org/slf4j/LoggerFactory.class
====

digitalbuddha:
I wanted to check as to anything else that is needed for this to get into a release version?

ceki:
@digitalbuddha @nitinverma Please confirm that https://github.com/qos-ch/slf4j/commit/fc0ad8652b8485175a6295648c05425284b8619a is satisfactory.

nitinverma:
@ceki Looks good to me. To be sure i'll patch the your commit to master and verify the performance improvement. I'll do that as soon as I get time within next 24 hours.

digitalbuddha:
howdy, is this making it into a release?

ceki:
@digitalbuddha  already part of the previous release.

PR #123, opened 2015-08-14
Remove unnecessary modifiers from interfaces

vbauer:


ceki:
Code in log4j-over-slf4j is legacy. There is no point in beautifying it.

vbauer:
NP

hazendaz:
If it's legacy why support it at all?  I agree with person making PR on this.  we still use this code in production.  There is a lot to be said about clean correct code.  It's not a pretty issue.  I don't see why a PR gets nixed for something like this.  Log4j legacy or not is still highly prevent in usage and slf4j supports it so why not have nice clean code that will pass all modern scanners.  There is no downside here.


Sent by Outlook for Android





On Fri, Oct 16, 2015 at 12:16 PM -0700, "Ceki Gulcu" <notifications@github.com> wrote:
Code in log4j-over-slf4j is legacy. There is no point in beautifying it.

---

PR #124, opened 2015-09-08
SLF4J-334: Update LoggerWrapper to pass throwable to bound logger in parameterized queries

millerw8:
Added logic in LoggerWrapper to send the ArgArray and Throwable parsed from the original parameters to the underlying logger.

Issue described here: http://jira.qos.ch/browse/SLF4J-334


amoosbr:
I also ran into this issue. Same i.e. trace log message with a exception as last parameter of multiple params results in different log entries, when using XLogger or Logger with i.e. log4j as logger.
As William already mentions in the bug descriptions are all loggers extending LoggerWrapper affected.

albertcsm:
I encountered the same issue and would really hope the the pull request got merged.

PR #125, opened 2015-09-29
SLF4J-335: jcl104-over-slf4j warning on release() should print to standard error

Andrew67:
This patch modifies the println calls in SLF4JLogFactory.release() to use System.err instead of System.out.


maswin:
When is this feature planned to be merged?

PR #126, opened 2015-10-07
Made MDC.MDCCloseable support nested try-with-resources blocks

sjoerdtalsma:
MDC.MDCCloseable.close() only could delete(key), now it can either put(key, previous) or delete(key).
Added unit tests in implementation modules where applicable to demonstrate this behavior.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -36,7 +36,7 @@ import org.slf4j.spi.MDCAdapter;
 /**
  * This class hides and serves as a substitute for the underlying logging
  * system's MDC implementation.
- * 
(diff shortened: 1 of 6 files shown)


ceki:
Adding a second parameter with "restore" capability seems to me surprising or non-obvious behavior. More importantly, the proposed changes would silently modify the behavior of existing code.

sjoerdtalsma:
We seem to differ in opinion about what is obvious in this case.
I respect your reluctance to modify existing behavior.
The change which I see as a fix is not critical to me, so no harm if you decide to reject it.
However, I would advise on using the added unit tests (modified to the current behavior) to be more explicit about the intended functionality.

sjoerdtalsma:
Closing this one, no activity for over a year.

PR #129, opened 2015-11-09
SLF4J-339: Embed log4j maven pom in log4j-over-slf4j

tmortagne:
SLF4J-340: Embed commons-logging maven pom in jcl-over-slf4j


ceki:
What kind of sorcery is this?

tmortagne:
See explanation in http://jira.qos.ch/browse/SLF4J-339.

ceki:
I've read the explanation you mention. It is still unclear why embedding a pom file helps.

tmortagne:
Because it is found in the JAR. I scan all JARs at startup to find Maven properties because that's the standard with JAR (which is a lot of jars including slf4j). So the idea here is to identify log4j-over-slf4j to be also a "log4j implementation" and can be used by anyone that needs log4j API.

ceki:
Si is this standard documented somewhere?

tmortagne:
I don't understand the question. What do you want documentation for ? The fact that Maven jar plugin automatically embbeds  Maven properties in the generated JAR ?

ceki:
I'd like documentation on the effects of the embedded pom.properties and pom.xml.

tmortagne:
On official Maven documentation I don't have anything more detailed than addMavenDescriptor property documentation on https://maven.apache.org/shared/maven-archiver/ and the comment in the source ():

    // We want to add the metadata for the project to the JAR in two forms:
    //
    // The first form is that of the POM itself. Applications that wish to
    // access the POM for an artifact using maven tools they can.
    //
    // The second form is that of a properties file containing the basic
    // top-level POM elements so that applications that wish to access
    // POM information without the use of maven tools can do so.

What I can say from experience is that
* it's always generated the same way
* it's designed to avoid collision (it's actually quite annoying to find them and I guess if the goal was to have only one by jar it would always be the same path)
* you often find several properties in standalone/big jars

ceki:
Which applications access the Maven properties files without Maven? I would like ascertain whether the embedding of the log4j properties and pom.xml files in log4j-over-slf4j.jar will break the build of an unsuspecting user.

tmortagne:
I guess things like Nexus and Artifactory provide some UI to upload an artifact and might be looking at this embedded pom.xml to avoid having to also upload a separated pom.xml. Now I don't think this is a real use case for something like slf4j which is on Maven central already.

Apart from this I don't really know anything else than XWiki (which is what I'm working on) that does that. Thing is Maven talk about "Applications that wish to access the POM for an artifact" but does not provide any kind of tool to do it (I guess this was the goal at some point but Maven is not as active as it used to be) so it's probably not done that much.

PR #130, opened 2015-12-01
Replaced maven-bundle-plugin with jar manifest template

mattbishop:
Picks up build-helper-maven-plugin config correctly.

### DIFF
--- a/osgi-over-slf4j/pom.xml
+++ b/osgi-over-slf4j/pom.xml
@@ -12,7 +12,7 @@
 
   <artifactId>osgi-over-slf4j</artifactId>
 
-  <packaging>bundle</packaging>
+  <packaging>jar</packaging>
   <name>OSGi LogService implemented over SLF4J</name>
 
   <url>http://www.slf4j.org</url>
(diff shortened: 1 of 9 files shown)


mattbishop:
The master build is not working. @ceki I'll rebase once you get that fixed.

mattbishop:
Fix for http://jira.qos.ch/browse/SLF4J-341

mattbishop:
I'd like to merge this now. Any objections?

hazendaz:
I don't use osgi but it looks clean and simple.  May I ask why the api piece is set to 1.6?


Sent from Outlook Mobile






On Thu, Dec 3, 2015 at 8:49 PM -0800, "Matt Bishop" <notifications@github.com> wrote:





I'd like to merge this now. Any objections?

---

ceki:
@hazendaz Because the 1.7.x series is binary compatible with the 1.6.x series. The only way for a 1.7.x series binding to pass a the compatibility check performed by 1.6.series slf4j-api.jar is for the former to ask for 1.6.99.

ceki:
I have no objections as long as the issue reporter, namely Antoine DESSAIGNE, is OK with it.

hazendaz:
Thanks for explaining that.


Sent from Outlook Mobile






On Fri, Dec 4, 2015 at 3:56 AM -0800, "Ceki Gulcu" <notifications@github.com> wrote:





@hazendaz Because the 1.7.x series is binary compatible with the 1.6.x series. The only way for a 1.7.x series binding to pass a the compatibility check performed by 1.6.series slf4j-api.jar is for the former to ask for 1.6.99.

---

mattbishop:
Hmm that is an artifact from the rebase I did from Ceki's revert. It
shouldn't be there but will vanish from history once merged as the hash of
the file will match master.

adessaigne:
Sorry about the delay.
I've checked out this PR and it seems to work in our project
There is one strange thing though : the ```Import-Package``` section in the ```MANIFEST.MF``` file has not 72 characters lines and 2 spaces at the beginning of the next line but 70 characters and a single space. We're using Apache Felix and it's working but other OSGi platforms might not like it.

mattbishop:
The line wrapping is performed by the jar plugin when it process the manifest. Also OSGi frameworks don't read this file directly, they use the JVM's classloader to read the entries into a Manifest class.

mattbishop:
Reporter is good with this change.

PR #132, opened 2015-12-09
Full implementation of OSGi Log Service specification (r4.2)

mattbishop:


ceki:
Given that there are very few osgi-over-slf4j users (less than 150), I am not keen in making further investments in osgi related logging in SLF4J.  May I suggest that the project be housed elsewhere?

mattbishop:
@ceki all right, good point. I'll look for a more OSGi-centric project to contribute it to.

ceki:
you can create an independent project on github. I'd be happy to link to it from the SLF4J web site.

PR #134, opened 2015-12-15
SLF4J-345: Move delegation check loop from log4j-over-slf4j.jar to slf4j-log4j12.jar

forsel:
Currently, the check on delegation loop for log4j is done in the log4j-over-sl4fj.jar, whereas the same check for jcl is done in the slf4j-jcl.jar. This causes different behaviour for log4j and jcl:
- For log4j, if both log4j-over-slf4j.jar and slf4j-log4j12.jar are present on the classpath, an IllegalStateException is always thrown, even if slf4j-log4j12.jar is not chosen as the actual binding.
- For jcl, if both jcl-over-slf4j.jar and slf4j-jcl.jar are present on the classpath, an IllegalStateException is ONLY thrown when slf4j-jcl.jar is chosen as the actual binding. If, for example, logback is also on the classpath and is chosen as the actual binding, then (JCLLoggerFactory class in) slf4j-jcl.jar is never initialized and therefore the delegation loop is not executed.

The log4j behaviour causes an issue for my situation, where I want to redirect log4j logging to slf4j/logback, but my classpath also contains (an older version of) slf4j-log4j12.jar. The actual binding is logback, so there is no real conflict, i.e. no danger of a delegation loop/stack overflow. Still, I'm unable to redirect log4j logging via slf4j/logback using log4j-over-slf4j.jar.

I propose to move the delegation loop check from log4j-over-slf4j.jar to slf4j-log4j12.jar, so that the delegation loop check behaviour for log4j will be the same as for jcl.

See also JIRA SLF4J-345.

### DIFF
--- a/log4j-over-slf4j/src/main/java/org/apache/log4j/Log4jLoggerFactory.java
+++ b/log4j-over-slf4j/src/main/java/org/apache/log4j/Log4jLoggerFactory.java
@@ -17,7 +17,6 @@
 package org.apache.log4j;
 
 import org.apache.log4j.spi.LoggerFactory;
-import org.slf4j.helpers.Util;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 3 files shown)


mattbishop:
I like the symmetry with JCL.

PR #138, opened 2015-12-30
Use getter methods for singleton instances in *Binder classes

rufoa:
This PR relates to JIRA issue [SLF4J-347](http://jira.qos.ch/browse/SLF4J-347)

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/impl/StaticMDCBinder.java
+++ b/slf4j-api/src/main/java/org/slf4j/impl/StaticMDCBinder.java
@@ -37,7 +37,16 @@ public class StaticMDCBinder {
     /**
      * The unique instance of this class.
      */
-    public static final StaticMDCBinder SINGLETON = new StaticMDCBinder();
+    private static final StaticMDCBinder SINGLETON = new StaticMDCBinder();
+
+    /**
+     * Return the singleton of this class.
+     * 
+     * @return the StaticMDCBinder singleton
+     */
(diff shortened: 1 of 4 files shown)


fzakaria:
+1

slf4j-api/src/main/java/org/slf4j/impl/StaticMDCBinder.java
```diff
@@ -37,7 +37,16 @@
     /**
      * The unique instance of this class.
      */
-    public static final StaticMDCBinder SINGLETON = new StaticMDCBinder();
```
mattbishop: I'm not sure about making this private right now. Perhaps mark it as @Deprecated instead so people know to switch to accessor method.

slf4j-api/src/main/java/org/slf4j/impl/StaticMarkerBinder.java
```diff
@@ -45,7 +45,16 @@
     /**
      * The unique instance of this class.
      */
-    public static final StaticMarkerBinder SINGLETON = new StaticMarkerBinder();
```
mattbishop: Same comment about deprecating the field.

rufoa:
Thanks @mattbishop - I've updated the PR

ceki:
Fixed by https://github.com/qos-ch/slf4j/commit/ea3cca72cd5a9329a06b788317a17e806ee8acd0

PR #140, opened 2016-02-18
Binding Ambiguity check skipped for Andriod

jeffreyzh:
As Dex cant by design have duplicate class, there would never be a 'Binding Ambiguity'.

1. LoggerFactory.getLogger is mostly called via MainActivity. Which gives a black screen before the first activity loads.
2. This operation takes 200-600 Inclusive CPU Time as per DDMS.

This fix get Incl CPU Time to sub 10

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -28,7 +28,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Enumeration;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


ceki:
@jeffreyzh Since 1.7.14, the binding ambiguity check is skipped the during initialization under Android in order to improve performance. What's going on?

jeffreyzh:
I update the version to 1.7.4 ，the following method takes a long time during initialization
```
loggerFactoryClassLoader.getResources(STATIC_LOGGER_BINDER_PATH)
```
the but I found the findPossibleStaticLoggerBinderPathSet is not skipped

ceki:
@jeffreyzh My apoligies, Problem fixed in https://github.com/qos-ch/slf4j/commit/5a95202d93c6994d
The fix will be part of version 1.7.17 to be released shortly.

jeffreyzh:
Well done，thank you so much.

PR #141, opened 2016-02-20
Changed cal10n dependency into an optional dependency

merusso:
When including slf4j-ext as a dependency, compiler warnings are thrown from transitive dependency cal10n. I'm hoping this is not a runtime dependency for slf4j, in which case it should not be included as a transitive dependency in projects that only wish to use slf4j-ext.


merusso:
Well, it didn't build. I guess it is a runtime dependency, bummer.

PR #144, opened 2016-02-29
fix scm url

oreissig:
Just a minor change: the SCM URL listed in the pom.xml is outdated. I changed it to point to the current github URL.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
   </licenses>
 
   <scm>
-    <url>https://github.com/ceki/slf4j</url>
+    <url>https://github.com/qos-ch/slf4j</url>
     <connection>git@github.com:qos-ch/slf4j.git</connection>
   </scm>  
 


oreissig:
I don't think the build failure is related to my change

```
multiThreadedInitialization(org.slf4j.impl.MultithreadedInitializationTest)  Time elapsed: 0.031 sec  <<< FAILURE!
java.lang.AssertionError: expected:<20> but was:<0>
	at org.junit.Assert.fail(Assert.java:93)
	at org.junit.Assert.failNotEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:128)
	at org.junit.Assert.assertEquals(Assert.java:472)
	at org.junit.Assert.assertEquals(Assert.java:456)
	at org.slf4j.impl.MultithreadedInitializationTest.multiThreadedInitialization(MultithreadedInitializationTest.java:84)
```

PR #154, opened 2016-06-17
SLF4J-236: Fix wrong marker checks in XLogger

alexeyr:
Fixes http://jira.qos.ch/browse/SLF4J-236.


alexeyr:
The test failure shouldn't happen, since the only changes are to `XLogger`. I've checked that the tests pass for me.

PR #155, opened 2016-07-19
Binding for edu.stanford.nlp.util.logging.Redwood

yfpeng:
Add "slf4j-redwood" to support Redwood Logger, which is used in Stanford
Natural Language toolkit.

https://github.com/gangeli/redwood

"Redwood was developed at Stanford by students in the Stanford NLP Group. A
tutorial in docs/tutorial.pdf, presented to the group, provides a good
introduction to the theory and use of Redwood. It also motivates what types
of projects would fit well into the paradigm."

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
     <module>slf4j-site</module>
     <module>slf4j-migrator</module>
     <module>slf4j-bom</module>
+    <module>slf4j-redwood</module>
   </modules>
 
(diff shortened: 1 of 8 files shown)


ceki:
I do not wish to merging into slf4j-redwood into slf4j proper as proposed in this PR. However, I'd be happy to add a link to the slf4j-redwood project on the slf4j web-site.

PR #157, opened 2016-08-04
Feature to allow suppressing unsuccessful init error message

hohwille:
As you can read in https://github.com/m-m-m/util/issues/194 there are cases where one uses slf4j in general library code that is also used in command-line-interface (CLI) programs.
In my case I want to ship the CLI library without any logger binding for slf4j in order to give users the freedom to choose on their own. However the CLI library already contains CLI programs by itself. If you want to run them without additional dependencies you get a warm welcome from slf4j saying:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

In case of a CLI program this output is really undesired.

I tried to be constructive and provide a minimal invasive "fix" for the problem that allows to suppress this output by setting a system property `slf4j.suppressInitError` (I tried to follow your current code and naming conventions) to true.

You can also add a public static non-arg method to `LoggerFactory` that sets the system property to true for you if you like. Also you can think of a system property that enforces nop logger no matter what is on your classpath. But as stated before I wanted to keep the change and impact minimal so I hopefully have a high change to get this PR accepted and you see this as a helpful improvement to slf4j. Thanks in advance.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -81,6 +81,7 @@ public final class LoggerFactory {
     static final String UNSUCCESSFUL_INIT_URL = CODES_PREFIX + "#unsuccessfulInit";
     static final String UNSUCCESSFUL_INIT_MSG = "org.slf4j.LoggerFactory in failed state. Original exception was thrown EARLIER. See also "
                     + UNSUCCESSFUL_INIT_URL;
+    static final String UNSUCCESSFUL_INIT_MSG_SUPPRESSION_SYSTEM_PROPERTY = "slf4j.suppressInitError";
 
     static final int UNINITIALIZED = 0;
     static final int ONGOING_INITIALIZATION = 1;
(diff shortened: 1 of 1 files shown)


hohwille:
completely redone as master changed and merge was not possible anymore.

LoneDev6:
Is this shit seriously been here since 2017?

hohwille:
No reaction after 6.5 years?

ceki:
Filing a jira issue would have helped this PR getting more attention as explained at the [project's home page.](https://github.com/qos-ch/slf4j#how-to-contribute-pull-requests).

As you write, the PR is non-invasive which is good. However, for a CLI, adding slf4j-nop would work as well. As such, I am closing this PR without merging.

hohwille:
> However, for a CLI, adding slf4j-nop would work as well. As such, I am closing this PR without merging.

This works if the users of the CLI library know about such things. However, users will try out the library and then see this SLF4J output messing the help page. Then they tell me that my library is shit. Unfortunately that is how life works. I can not include `slf4j-nop` myself as then other users that want to use a real logger writing to a file while the CLI is only producing basic output.

ceki:
How about including slf4j-simple which Maven uses? It is still quite configurable.

PR #159, opened 2016-08-26
Adding a bridge for Twitter-Logging to SLF4j.

Dirk-c-Walter:
It's mostly the same as jul but with a few tweaks.

The jul bridge works for sending logs over, but this one has the changes needed to support their
admin tools.

Accordingly the code is mostly taken from the jul bridge, I'm just pushing it out here so it can be of use to other people. (and selfishly so it will just show up in my repositories whenever a new version is released.


devshorts:
Whats the status on this PR? I'd love to have this

PR #161, opened 2016-10-24
Active logger factory can be overridden with a custom instance

bbobcik:
Implements a method for bypassing logger factory autodiscovery mechanism. A specific logger factory instance is registered within LoggerFactory and then used directly without invoking binding phase.
Related JIRA: [http://jira.qos.ch/browse/SLF4J-378](http://jira.qos.ch/browse/SLF4J-378)

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -91,6 +91,7 @@ public final class LoggerFactory {
     static final String JAVA_VENDOR_PROPERTY = "java.vendor.url";
 
     static boolean DETECT_LOGGER_NAME_MISMATCH = Util.safeGetBooleanSystemProperty(DETECT_LOGGER_NAME_MISMATCH_PROPERTY);
+    static ILoggerFactory FACTORY_OVERRIDE;
 
     /**
      * It is LoggerFactory's responsibility to track version changes and manage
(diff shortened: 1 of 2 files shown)


yairogen:
@ceki can you update status on this PR?
This PR is approved and merged can help me use logback with my own logger implementation. Today I have to create my own logback binder which collides with the one in logback-classic.
On that note: why not offer binder in dedicated jar allowing users to easily supply their own binder?

ceki:
The current binding rules are stupid and unflexible for a good reason. Indeed, they can be understood by any developer within minutes and that developer can be expected to reason correctly about them.

The setLoggerFactory() mechanism proposed here is ad hoc and, dare I say, unprincipled. 

In any case, there is no way new binding rules will/can be introduced in the 1.7.x series.

FYI, the 1.8 branch will use the ServiceLoader mechanism with similar unflexible logic as in the 1.7.x. If you want to see the rules change, then I invite you to try your luck on the 1.8 branch. However, setLoggerFactory()'s open-buffet approach is very likely to receive very little love on the 1.8 branch as well. 

On the 1.8 branch, the first available service provider is selected. However, the selection mechanism can be "enhanced" with different procider selection rules, e.g. by priority...

ceki:
Since this PR is unlikely to be merged, I propose to close and continut the discussion on the slf4j-dev mailing list or alternatively on  http://jira.qos.ch/browse/SLF4J-378

PR #166, opened 2016-12-16
Only remove Throwable if unconsumed by pattern.

huxi:
This is a preparation for a [LOGBACK-1183](http://jira.qos.ch/browse/LOGBACK-1183) fix while keeping the optimizations of [LOGBACK-873](http://jira.qos.ch/browse/LOGBACK-873) in place.

It replaces the package-private method `Throwable getThrowableCandidate(Object[] argArray)` with a public `Throwable extractUnconsumedThrowable(String messagePattern, Object[] argArray)` method.

The unit test that was actually verifying the wrong behavior before has been changed and extended for both cases:
- a `Throwable` at the end of the argument array is removed if is **not** "consumed" by the message pattern.
- a `Throwable` at the end of the argument array is left intact if it **is** "consumed" by the message pattern.

"consumed" means that a placeholder in the message pattern exists that will be replaced by the `Throwable`. If we assume an argument array `{"bar", new Throwable()}` then the pattern `"foo {} {}"` **would** "consume" the `Throwable` resulting in the formatted string `"foo bar java.lang.Throwable"` while the pattern `"foo {}"` **would not** consume the `Throwable` resulting in the formatted string `"foo bar"`. This was the behavior of Logback below 1.1.

The `extractUnconsumedThrowable` method would then by used by Logback instead of `EventArgUtil.extractThrowable`.

I initially tried to fix [LOGBACK-1183](http://jira.qos.ch/browse/LOGBACK-1183) solely in Logback but had to realize that SLF4J would later remove the `Throwable` anyway while formatting the message, regardless of whether or not `EventArgUtil.extractThrowable` left it intact in the argument array.

`LoggingEvent` in Logback would then see the following change:
**Original code:**
```java
    private Throwable extractThrowableAndRearrangeArguments(Object[] argArray) {
        Throwable extractedThrowable = EventArgUtil.extractThrowable(argArray);
        if (EventArgUtil.successfulExtraction(extractedThrowable)) {
            this.argumentArray = EventArgUtil.trimmedCopy(argArray);
        }
        return extractedThrowable;
    }
```

**New code:**
```java
    private Throwable extractThrowableAndRearrangeArguments(String messagePattern, Object[] argArray) {
        Throwable extractedThrowable = MessageFormatter.extractUnconsumedThrowable(messagePattern, argArray);
        if (EventArgUtil.successfulExtraction(extractedThrowable)) {
            this.argumentArray = EventArgUtil.trimmedCopy(argArray);
        }
        return extractedThrowable;
    }
```

This private method is only called in the constructor.
**Original code:**
```java
        if (throwable == null) {
            throwable = extractThrowableAndRearrangeArguments(argArray);
        }
```

**New code:**
```java
        if (throwable == null) {
            throwable = extractThrowableAndRearrangeArguments(message, argArray);
        }
```

Applying those two changes in Logback would fix [LOGBACK-1183](http://jira.qos.ch/browse/LOGBACK-1183).

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/helpers/MessageFormatter.java
+++ b/slf4j-api/src/main/java/org/slf4j/helpers/MessageFormatter.java
@@ -163,6 +163,20 @@ final public class MessageFormatter {
             return 0;
         }
 
+        if(-1 == messagePattern.indexOf(DELIM_START))
+        {
+            // Special case: no placeholders at all.
+
+            // This is an optimization because charAt checks bounds for every
+            // single call while indexOf(char) isn't.
+
+            // Big messages without placeholders will benefit from this shortcut.
+
+            // the result of indexOf can't be used as start index in the loop
+            // below because it could still be escaped.
+            return 0;
+        }
+
         int result = 0;
         boolean isEscaped = false;
         for(int i = 0; i < messagePattern.length(); i++) {
--- a/slf4j-api/src/main/java/org/slf4j/helpers/MessageFormatter.java
+++ b/slf4j-api/src/main/java/org/slf4j/helpers/MessageFormatter.java
@@ -152,20 +152,60 @@ final public class MessageFormatter {
     }
 
 
-    static final Throwable getThrowableCandidate(Object[] argArray) {
+    /**
+     * Counts the number of unescaped placeholders in the given messagePattern.
+     *
+     * @param messagePattern the message pattern to be analyzed.
+     * @return the number of unescaped placeholders.
+     */
+    private static int countArgumentPlaceholders(String messagePattern) {
+        if(messagePattern == null) {
+            return 0;
+        }
+
+        int result = 0;
+        boolean isEscaped = false;
+        for(int i = 0; i < messagePattern.length(); i++) {
+            char curChar = messagePattern.charAt(i);
+            if(curChar == ESCAPE_CHAR) {
+                isEscaped = !isEscaped;
+            }
+            else if(curChar == DELIM_START) {
+                if(!isEscaped) {
+                    if(i < messagePattern.length() - 1) {
+                        if(messagePattern.charAt(i + 1) == DELIM_STOP) {
+                            result++;
+                            i++;
+                        }
+                    }
+                }
+                isEscaped = false;
+            }
+            else {
+                isEscaped = false;
+            }
+        }
+        return result;
+    }
+
+
+    public static Throwable extractUnconsumedThrowable(String messagePattern, Object[] argArray) {
         if (argArray == null || argArray.length == 0) {
             return null;
         }
 
         final Object lastEntry = argArray[argArray.length - 1];
         if (lastEntry instanceof Throwable) {
-            return (Throwable) lastEntry;
+            int argumentCount = countArgumentPlaceholders(messagePattern);
+            if(argumentCount < argArray.length) {
+                return (Throwable) lastEntry;
+            }
         }
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


ceki:
@huxi It is clear that you have given this more thought than I have. 

I was under the impression that it was generally OK to consume the throwable  if it's the last argument regardless of the {} placeholders in the pattern. There have been no complaints about this approach until LOGBACK-1183.  

Do we really want to support the case covered by LoggerMessageFormattingRegressionTest.formattingRegression ?

huxi:
@ceki 
It was supported previously, is supported in log4j2 and this regression actually bit me in production.

Counting the tokens is seriously fast, i.e. in no way comparable to the performance impact of actually formatting a message.

If you want to optimize message formatting further then try to get rid of `trimmedCopy` because that causes unnecessary GC pressure.

I re-implemented the SLF4J `MessageFormatter` in [MessageFormatter](https://github.com/huxi/lilith/blob/master/lilith-data/logging/src/main/java/de/huxhorn/lilith/data/logging/MessageFormatter.java) and [SafeString](https://github.com/huxi/sulky/blob/master/sulky-formatting/src/main/java/de/huxhorn/sulky/formatting/SafeString.java) with enhanced abilities (error resilient, better formatting style than Java (e.g. `byte[]` formatted as hex, `null` can be distinguished from `"null"`, empty `Collection<String>` can be distinguished from `Collection<String>` containing **only** a single empty String), etc.) and exhaustive tests.

I'm 100% OK if you want to "steal" any or all of that code. `java.time` support requires Java 8 so that would have to be removed and formatting of `java.util.Date` to UTC-ISO8601 String would hav to be re-implemented differently, for example.

gagoman:
Agreement was reached that this issue (https://github.com/qos-ch/logback/pull/374) need to be fixed here.

gagoman:
@ceki any help is required with this? I've also faced an issue with throwable trimmed by `MessageFormatter`

fenik17:
@ceki i'm agree that it's a regression, and needs to be fixed.

> I was under the impression that it was generally OK to consume the throwable if it's the last argument regardless of the {} placeholders in the pattern

If someone uses a placeholder in a template, he expects that placeholder will be replaced with text...

@huxi approach from your patch has a big CPU overhead due to a double parse of the `messagePattern`. I think that it's enough at the end of the `arrayFormat()` method to check the count of replaced placeholders (`L` variable). Something like this:
```java
        if (L < argArray.length) {
            argArray = trimmedCopy(argArray); 
        }
```

huxi:
@fenik17 I'm not sure what you mean about the big CPU overhead. I was aiming for a minimal change and counting is only performed if the last entry is actually a `Throwable`.

I think it's more questionable if trimming the array (i.e. creating a copy of the original array minus the last entry) is necessary at all - but I wanted to keep the scope of the PR manageable.

gagoman:
Can I provide any help with this issue?

jezovuk:
I was hit by this today (via Netty, in their `AbstractChannelHandlerContext`). Any chance of releasing the fix soon?

ceki:
@jezovuk  Can you please provide a reference to the code of AbstractChannelHandlerContext?

komu:
I got a bug report from a user of my library saying that exceptions are not printed correctly and after some investigation 
found this bug. The offending line in my code was:

```java
log.warn("rolled back transaction because of exception: {}", e, e); 
```

Changing the code to the following fixes the problem, but feels like a silly thing to do:

```java
log.warn("rolled back transaction because of exception: {}", e.toString(), e); 
```

Surely it might be better to not include the exception in the actual message and leave it to the format to decide, but it's something that people do anyway.

PR #167, opened 2016-12-17
SLF4J-154 Inferring logger from calling class.

thorntonrp:
Added an operation to get logger inferred from the calling class. The implementation uses the same utility that is used to detect a logger name mismatch, resulting in no extra overhead.

Tests included.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -392,17 +392,18 @@ public final class LoggerFactory {
         return logger;
     }
 
-    public static Logger getLogger()
-    {
-        StackTraceElement[] stack = new Throwable().getStackTrace();
-        for(StackTraceElement item : stack) {
-            String className = item.getClassName().split("\\$")[0];
-            // Exclude our own class and any proxies
-            if(!className.equals(LoggerFactory.class.getName()) && !className.equals("com.sun.proxy")) {
-                return getLogger(className);
-            }
+    /**
+     * Return a logger named corresponding to the class of the caller,
+     * using the statically bound {@link ILoggerFactory} instance.
+     *
+     * @return logger
+     */
+    public static Logger getLogger() { // SLF4J-154
+        Class<?> autoComputedCallingClass = Util.getCallingClass();
+        if (autoComputedCallingClass != null) {
+            return getLogger(Util.getCallingClass().getName());
         }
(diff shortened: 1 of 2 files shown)


huxi:
You should replace
```java
String className = callingClass.getName().split("\\$")[0];
```
with
```java
String className = callingClass.getName();
int dollarIndex = className.indexOf('$');
if(dollarIndex > -1) {
    className = className.substring(0, dollarIndex);
}
```

It's way faster.

slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
```diff
@@ -392,6 +401,20 @@ public static Logger getLogger(Class<?> clazz) {
         return logger;
     }
 
+    public static Logger getLogger()
+    {
+        Class<?>[] callingClasses = CONTEXT.getCallingClasses();
+        for (Class<?> callingClass : callingClasses) {
+            String className = callingClass.getName().split("\\$")[0];
+            // Exclude our own class and any proxies
+            if (!className.equals(LoggerFactory.class.getName()) && !className.equals("com.sun.proxy")) {
```
huxi: I think this should be `!className.startsWith("com.sun.proxy")`.

thorntonrp:
This is the final version at 363afac:

```java
    public static Logger getLogger() { // SLF4J-154
        Class<?> autoComputedCallingClass = Util.getCallingClass();
        if (autoComputedCallingClass != null) {
            return getLogger(Util.getCallingClass().getName());
        }
        Util.report("Failed to detect logger name from caller.");
        return getLogger(Logger.ROOT_LOGGER_NAME);
    }
```

thomasmueller:
> why not reuse the variable?

I would say, because of the custom warning "Failed to detect logger name from caller.". But getCallingClass() should only be called once (untested):

        public static Logger getLogger() { // SLF4J-154
            String name;
            Class<?> autoComputedCallingClass = Util.getCallingClass();
            if (autoComputedCallingClass == null) {
                Util.report("Failed to detect logger name from caller.");
                name = Logger.ROOT_LOGGER_NAME;
            } else {
                name = autoComputedCallingClass.getName();
            }
            return getLogger(name);
        }

adorokhine:
What's the difference between this pull request and https://github.com/qos-ch/slf4j/pull/29 ?

thorntonrp:
It looks like this PR has the same goal as #29

PR #171, opened 2017-02-19
Several improvements in MessageFormatter

fenik17:
- avoid redundant instantiations;
- replace HashMap to IdentityHashMap;
- minor optimization and simplification;
- some new tests.


fenik17:
Could not show perf impact with benchmarks.

PR #172, opened 2017-02-23
Add lifecycle-mapping ignore for plugin execution

hazendaz:
Reopening this as solution I provided is a little bit smarter.  The m2e.version only exists when you are sitting in eclipse.  Not actually at all when it runs.  So what that does is still cause eclipse to not show a warning.  This way while it states it has no harm elsewhere, it's not even a factor at all during maven run in any way.  I didn't even realize myself how exactly m2e.version property came into play assuming that it actually was part of the build when running inside eclipse but that is not the case.  It's purely how m2e, the pom, and eclipse interact with each other before any physical build request is ever made.  Hopefuly that helps with some insite there.

### DIFF
--- a/slf4j-api/pom.xml
+++ b/slf4j-api/pom.xml
@@ -81,40 +81,47 @@
         </configuration>
       </plugin>
     </plugins>
-
-    <pluginManagement>
-    	<plugins>
-    		<!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
-    		<plugin>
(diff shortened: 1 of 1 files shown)


ceki:
I understand that the ''m2e.version" check disables lifecycle-mapping exclusion of maven-antrun-plugin. However, I don't see the added value of this check.

hazendaz:
Fair enough.  Someone asked I do that on another repo long ago.  I forget why but have done it ever since.  Looking at it I cannot make a compelling real life reason so I'm good with your decision.  Thanks.

PR #174, opened 2017-03-07
Add a high service ranking to the registered LogService

msqr:
This PR adds a high service ranking to the registered LogService so it will take precedent over any built-in LogService provided by the OSGi runtime (e.g. Equinox). I figured if deploying the slf4j LogService most people would expect this implementation to be used, but that Equinox's built-in service is registered as well at the default service level so bundles importing LogService might get that one instead of the slf4j implementation.


ceki:
Please create a Jira issue for this at jira.qos.ch

msqr:
Done: https://jira.qos.ch/browse/SLF4J-398

PR #175, opened 2017-03-16
Ignore errors that can occur when converting a message to a string

thsnoopy:
An exception may occur when converting a message to a string.
For example, someone's incorrect implementation in toString method.
I think it is better to ignore the exception and let the Category(Logger) continue to run.
What do you think about it?


ceki:
Please link this PR to a Jira issue.

thsnoopy:
@ceki 
I've created Jira issue and also remove Mockito :)
- Jira : https://github.com/qos-ch/slf4j/pull/175

PR #176, opened 2017-05-24
Try to read logger properties from env variables

kmakowski:
Try to read logger properties from env variables when values are not present in system properties.

### DIFF
--- a/slf4j-simple/src/main/java/org/slf4j/impl/SimpleLoggerConfiguration.java
+++ b/slf4j-simple/src/main/java/org/slf4j/impl/SimpleLoggerConfiguration.java
@@ -136,9 +136,15 @@ public class SimpleLoggerConfiguration {
         String prop = null;
         try {
             prop = System.getProperty(name);
+            
+            if (prop == null) {
+                prop = System.getenv(name);
+            }
+            
         } catch (SecurityException e) {
             ; // Ignore
         }
+        
         return (prop == null) ? properties.getProperty(name) : prop;
     }
 


ceki:
Please create a jira issue for this PR

kmakowski:
https://jira.qos.ch/browse/SLF4J-410

slf4j-simple/src/main/java/org/slf4j/impl/SimpleLoggerConfiguration.java
```diff
@@ -136,9 +136,15 @@ String getStringProperty(String name) {
         String prop = null;
         try {
             prop = System.getProperty(name);
+            
+            if (prop == null) {
+                prop = System.getenv(name);
```
sdavids: Under a SecurityManger this might throw an AccessControlException

sdavids: `System.getProperty` too by the way

slf4j-simple/src/main/java/org/slf4j/impl/SimpleLoggerConfiguration.java
```diff
@@ -136,9 +136,15 @@ String getStringProperty(String name) {
         String prop = null;
         try {
             prop = System.getProperty(name);
+            
+            if (prop == null) {
```
sdavids: This should be in its own try-catch block:

One might not have the permission to access `System.getProperty` but have the permission to access `System.getenv`

sdavids:
What Spring Boot does:

> Environment variables are bound by lowercasing and replacing _ with . .
https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#environment-variables

Might be overkill for SLF4J though ...

sdavids:
Unfortunately we have properties like:

`slf4j.detectLoggerNameMismatch`

One could only use a case-insensitive compare in that case.

sdavids:
I am unsure if updating the Chinese readme fixes this issue …

Please reopen.

PR #178, opened 2017-06-19
Fix the SCM connection URL in pom.xml

sschuberth:
The "connection" should be an URL valid for read-only access and has to follow
the syntax as described at https://maven.apache.org/pom.html#SCM.


sschuberth:
@ceki, any chance go get this trivial fix merged before the next release?

PR #184, opened 2017-10-09
Add Automatic-Module-Name to MANIFEST

jodastephen:
The idea of this commit is to get a release 1.7.26 done with automatic module names for the four outward-facing modules. This allows the issues of service loader, module-info and so on to be sidestepped for now.

**This PR is not intended to be merged onto the master branch.** Instead, it is intended for there to be a new branch in slf4j starting at 1232e8529f5844e03e4d528a88185413761853e2 where this is the first commit on the branch. I can't get GitHub to do this automatically, so you'd need to create the branch in slf4j first, and I can then re-raise the PR against the new branch. Alternatively, you can cherry pick this commit onto the branch.

There is a case to add `Automatic-Module-Name` to the inward-facing (impl) jars, but I'm still working out what the best strategy for those is.

thanks

### DIFF
--- a/jcl-over-slf4j/pom.xml
+++ b/jcl-over-slf4j/pom.xml
@@ -28,4 +28,31 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <!-- This module replaces the real commons-logging, thus it must have the same module name -->
+      <!-- http://svn.apache.org/repos/asf/commons/proper/logging/trunk/pom.xml -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


orionll:
Any progress on this merge? It would be nice to have SLF4j 1.8 with right Automatic-Module-Name

varunnvs92:
Any plans on merging this change?

varunnvs92:
@ceki 
Any update on https://jira.qos.ch/browse/SLF4J-420?

What are you timelines on launching a version with JDK 9 module support?

jodastephen:
Mostly fixed by https://github.com/qos-ch/slf4j/commit/bf50c49d10e37e4d0c82a2529c2f074237262db1

PR #185, opened 2017-10-17
Documentation change

bjornblomqvist:
Documented that if the last argument given is an exception it will
be printed in the log. As this is not documented its unclear how
to use arguments and include an exception.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/Logger.java
+++ b/slf4j-api/src/main/java/org/slf4j/Logger.java
@@ -99,6 +99,8 @@ public interface Logger {
      * <p/>
      * <p>This form avoids superfluous object creation when the logger
      * is disabled for the TRACE level. </p>
+     * <p>In most implementations if the the last argument given is an exception
+     * it will be handled the same way as in {@link #trace(String, Throwable)}.<p/>
      *
      * @param format the format string
      * @param arg    the argument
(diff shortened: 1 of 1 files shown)


ceki:
Instead of repeating the same text multiple times, I rather the statement "Note that logging statements can be parameterized in presence of an exception/throwable" be emphasized or expanded upon.

aglassman:
I agree that the JavaDoc should be more verbose here.  It is unclear from the JavaDoc that the last parameter in the array is special.  This is hidden functionality IMO.  JavaDoc on the interface should explicitly state that if the last element of the array extends Throwable, it will not be used to template the string, and will be "the exception (throwable) to log". 

Documenting this expected functionality in docs outside of the JavaDoc is fine, but it should explicitly be stated in the interface contract.

ceki:
Please create a Jira issue linking to this PR.

bjornblomqvist:
@ceki I am not able to reach https://jira.qos.ch/ It looks to be down. Could you please have a look at it?

bjornblomqvist:
@aglassman where is this documented? I only found this by thinking that it should be there and reading the source code.

aglassman:
@bjornblomqvist I was only able to find a reference to this functionality on the slf4j FAQ page.

aglassman:
Created: https://jira.qos.ch/browse/SLF4J-446

alb-i986:
+1

I was about to open a jira but I found the jira related to this PR.

@ceki Any chance it can make it into master?

Right now it's really not adhering to the principle of least astonishment.
The thing is that there are some methods which _do_ take a Throwable, and others which don't.
So I will expect the methods not taking a Throwable to not support it, _unless_ stated otherwise in the javadoc.

I see that there's a line in the javadoc at the _class level_ mentioning the thing, but personally I had completely missed that.
https://github.com/qos-ch/slf4j/blob/e2c29c81df6fb405a3c5a6dfd1cf28ba26a7c1d6/slf4j-api/src/main/java/org/slf4j/Logger.java#L73
I wouldn't have missed it if it were at the method level, as this PR is proposing

PR #187, opened 2017-11-11
A few toolchains etc. tweaks to try and get Travis build working

javabrett:
Travis build on master has been down for a bit.  This change:

- Sets Travis `dist` to `trusty`.
- Requires `oraclejdk9` in Travis because we now depend on it.
- Disables `maven-toolchains-plugin` when running in Travis, they won't support it, see travis-ci/travis-ci#2727 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
+dist: trusty
 language: java
+jdk:
+  - oraclejdk9
 notifications:
   email:
       - notification@qos.ch
--- a/pom.xml
+++ b/pom.xml
@@ -157,6 +157,7 @@
         <version>1.1</version>
         <executions>
           <execution>
+            <id>toolchain</id>
             <goals>
               <goal>toolchain</goal>
             </goals>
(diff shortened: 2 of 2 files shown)


javabrett:
Any feedback on this effort to make the Travis build succeed fir `master`?

javabrett:
Raised https://jira.qos.ch/browse/SLF4J-427 .

ceki:
Up until a few weeks ago, logback was built using Java 9 (without toolchains) with Java 7 as the target JDK.  While compilation, testing and artifact creation works under this structure, javadocs cannot be produced under Java 9. In turn, this precludes releases on Maven central (because it checks whether the release has javadoc artifacts).

At present time, the only way to make a valid built is with toolchains where Java7 is the default. If understand correctly, Travis does not support toolchains. While I care about Travis, I care a lot more about having valid/releasable builds.

ceki:
Doh. Your idea is to disable toolchains for Travis.

javabrett:
Yes this is primarily about Travis builds when `env.TRAVIS` is set, activating that profile.  It is a quick-and-dirty fix that should only impact the Travis build, which whilst not critical, is a nice-to-have and helps identify build-failures and problems with PRs.

It is possible that the Travis build can be further improved with some matrix config to handle the different JDK requirements.  This only attempts to put it back to passing in the first instance.

ceki:
I had to inverse Java 1.6/Java 9 declaration in the <jdk> element in <toolchains>. This allows javadoc to be built. Apparently, the last jdk declaration becomes the default for other plugins.

javabrett:
> I had to inverse Java 1.6/Java 9 declaration in the element in . This allows javadoc to be built. Apparently, the last jdk declaration becomes the default for other plugins.

Ah good catch, I would not have expected that.

> And Travis build passes as well. :-)

'Tis good to see, thanks!

PR #189, opened 2017-11-18
Supplier added to logger interface

nikita-sheremet-java-developer:
The java 8 `Supplier<String>` is added to Logger interface.

Inspired from [kotlin logger](https://github.com/MicroUtils/kotlin-logging)

Some points to pay attention:

1. Could you please look at `org.slf4j.event.EventRecodingLogger#recordEvent(org.slf4j.event.Level, org.slf4j.Marker, java.util.function.Supplier<java.lang.String>, java.lang.Throwable)` method (TODO added). It gets message from supplier without checking log level. This may leads to performance issues. It is unclear (at least for me) how is `org.slf4j.event.SubstituteLoggingEvent` handled. If it is just a class in same jvm so we can add a `.setSupplier` method to it. But if this class is plain dto and it can be serializsed adding supplier method is not a solution. Could you please review this point?
2. `org.slf4j.jul.JDK14LoggerAdapter.java`
At first glace this class should not have any `Supplier` methods. But I changed it in the same way as other classes because minimal jdk version was `1.6`, so `JDK14LoggerAdapter` is not restricted to `1.4`. But I can be wrong. Could you please check this? If adding `Supplier<String>` is not a solution should `JDK14LoggerAdapter` be deleted?
3. I have made javadocs for new methods similar as others, but english is not my native language - could you please check javadocs?

I saw that this project is not being developed intensively and most of PR about comments to javadocs or typos and so on. But this changes about API and if you provide feedback quickly and I quickly make new changes -> new slf4j API will be quickly published :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
         <configuration>
           <toolchains>
             <jdk>
-              <version>1.9</version>
+              <version>9</version>
             </jdk>
           </toolchains>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
         <configuration>
           <toolchains>
             <jdk>
-              <version>1.8</version>
+              <version>8</version>
             </jdk>
           </toolchains>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -35,14 +35,14 @@
   <properties>
     <stable.version>1.7.25</stable.version>
     <!-- java.util.ServiceLoader requires Java 6 -->
-    <jdk.version>1.6</jdk.version>
+    <jdk.version>1.8</jdk.version>
     <maven.compiler.source>${jdk.version}</maven.compiler.source>
     <maven.compiler.target>${jdk.version}</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <!-- used in integration testing -->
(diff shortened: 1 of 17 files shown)


nikita-sheremet-java-developer:
I have tried jdks:

1. 1.8
2. 1.9
3. 8
4. 9

But travis still couldn't build the project. Could you please new jdk into travis?

ceki:
We are not yet ready to move the slf4j-api to Java 8.

nikita-sheremet-java-developer:
Sad. :( But:

1. Why or can I help with moving to java 8?
2. When do you planned to so (I saw moudle-info file which are support by java 9 or later)?

ceki:
- The slf4j-api is a common denominator in thousands if not millions of projects. It follows that in order to support this large number of projects, we still need to support Java 6. 
- Java 9 support retains Java 6 compatibility.

hazendaz:
With exception to many plugins such as animal sniffer that crash due to java 9 features.  While fixed on newer versions I suspect adding java 9 features is going to cause users pain and if this isn't raising java version it's rather pointless.

nikita-sheremet-java-developer:
> The slf4j-api is a common denominator in thousands if not millions of projects. It follows that in order to support this large number of projects, we still need to support Java 6.

I understand, but `slf4j` have to move to new java versions in future. Theoretically slf4j with java 8 api may have another version like `2.0.0`, `1.8-java8` or `java-8-1.8`. So all projects which support java 8 can move, all other can stay with old versions. From other side it is needed to support 2 projects (branches) instead of 1.

Have you considered such approach or you think that it will be difficult to manage/support both versions (branches)?

ceki:
Parallel versions is defintely a no go.

PR #191, opened 2017-12-06
SLF4J-422 Include unused arguments.

epabst:


epabst:
The build failed for an unrelated reason: 
Cannot find matching toolchain definitions for the following toolchain types: jdk [ version='1.6' ]

PR #193, opened 2018-02-03
fix for SLF4J-429

pa314159:


ceki:
It is not possible to log at the OFF level. So the "mapping" of the OFF level does not matter. Please create a unit test illustrating the problem.

pa314159:
It is possible to log at any level by calling java.util.logging.Logger.log, which is what actually Jersey or any other application that configures the level at runtime does.

See your test SLF4JBridgeHandlerTest.testLevels failing here https://github.com/pa314159/slf4j/tree/SLF4J-429/failure

PR #197, opened 2018-05-06
Added toolchain setup instructions

toolforger:


toolforger:
This has been sitting unreviewed for 10 days.
Sinking a few hours into an update and not getting feedback is unrewarding.
Starting work on something and having to shelve it for 10 days starts swapping out memories, so it's an ineffective workflow.

For these reasons I conclude:
Please feel free to merge or reject, but I don't care about this PR anymore.
(Actually I won't try to contribute to SLF4J again.)

PR #198, opened 2018-05-18
Make slf4j with java.util.ServiceLoader properly work in OSGi by using the Service Loader Mediator Specification

Vampire:
As a workaround one can create a fragment for the slf4j-api bundle, e. g. with Gradle using

```groovy
plugins {
    id 'osgi'
}

task slf4jApiFragmentJar(type: Jar) {
    baseName = project.name + '-slf4jApiFragment'
    manifest = osgiManifest {
        name = 'net.kautler.test.osgi.slf4j.api.fragment'
        symbolicName = 'net.kautler.test.osgi.slf4j.api.fragment'
        version = '1.0.0'
        classesDir = temporaryDir
        classpath = files()
        instruction 'Fragment-Host', 'slf4j.api'
        instruction 'Require-Capability', '' +
                'osgi.extender;' +
                'filter:="(osgi.extender=osgi.serviceloader.processor)",' +
                'osgi.serviceloader;' +
                'filter:="(osgi.serviceloader=org.slf4j.spi.SLF4JServiceProvider)"'
    }
}
```

### DIFF
--- a/slf4j-api/src/main/resources/META-INF/MANIFEST.MF
+++ b/slf4j-api/src/main/resources/META-INF/MANIFEST.MF
@@ -9,3 +9,6 @@ Export-Package: org.slf4j;version=${parsedVersion.osgiVersion},
   org.slf4j.helpers;version=${parsedVersion.osgiVersion},
   org.slf4j.event;version=${parsedVersion.osgiVersion}
 Import-Package: org.slf4j.spi;version=${parsedVersion.osgiVersion}
+Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.service
+ loader.processor)",osgi.serviceloader;filter:="(osgi.serviceloader=or
+ g.slf4j.spi.SLF4JServiceProvider)"


Vampire:
Of course it would be nicer if the native OSGi service loader mechanism would be supported, but that would need a bit more work I think. (and more experience with OSGi than I have currently)

rgansevles:
This pull-request is related to the fix in log4j2 see https://issues.apache.org/jira/browse/LOG4J2-2305

On the log4j-side  a fix has been applied and has been verified icw this pull-request.

Please consider this pull-request for next release.

HennoVermeulen:
I experienced the same problem trying to bind slf4j 1.8 to log4j1 in an OSGi environment using bundles slf4j.api (version 1.8.0.alpha2) and slf4j.log4j12 (version 1.8.0.alpha2). I can solve it by only changing the manifest of slf4j.log4j12 to include "<Fragment-Host>slf4j.api</Fragment-Host>" so that slf4j.api can find the required resources from its classpath.

As slf4j-log4j12 is part of slf4j itself, could the fix for log4j1 be included as part of this pull request? The manifest file is [here](https://github.com/qos-ch/slf4j/blob/1_8_0-SNAPSHOT/slf4j-log4j12/src/main/resources/META-INF/MANIFEST.MF)

ceki:
a jira ticket would help me track the issue. Can you please create a jira ticket?

Vampire:
Sure: https://jira.qos.ch/browse/SLF4J-457

jcompagner:
Are there any plans to put this in? 1.8 release was all about service loader and so on
and this is a another small step so that also in osgi containers this service loader can be used..

ceki:
The conversation in LOG4J2-2305 seemed quite involved. I would appreciate if @Vampire or @jcompagner  tested the commit and confirmed that it was working as expected.

rgansevles:
I can confirm is works as expected.

Will there be a 1.8 new beta and if yes, can you also merge this fix there?
Problem now is that log4j had a special jar to deal with slf4j-1.8 which does not accept slf4j-2, so for us to work with this we need a change to log4j.

Rob

ceki:
@rgansevles  Thank you for the confirmation.

The SLF4J 1.8 line has been subsumed by 2.0 which builds on 1.8 but adding a fluent API.  The fluent API is transparent to existing implementations unless they wish to support it natively.

I suppose log4j's special jar specifies a version number for slf4j-api. Right? 

It should be noted that for backends such log4j2, there might be non-transparent API changes in SLF4J 2.0-alpha* before going beta.

PR #200, opened 2018-10-10
Update AppenderSkeleton to support Jalopy implementation

stefanleh:
Add protected String `name` for backward compatibility with Jalopy 1.5-rc3. 
Jalopy sets the name directly instead using the designated setter method.


stefanleh:
Ping.

PR #203, opened 2019-02-20
Starting 2.0

ffissore:
This PR sets the minimum java version to 1.8 and includes slf4j-fluent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 dist: trusty
+
 language: java
+
 jdk:
   - oraclejdk11
 
@@ -7,5 +9,4 @@ notifications:
   email:
       - notification@qos.ch
 
-script: clean install
   
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,9 @@ jdk:
 
 notifications:
   email:
-      - notification@qos.ch
+      - notification@qos.c
+
+install: /bin/true
 
 script:
   - mvn clean
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ dist: trusty
 language: java
 jdk:
   - oraclejdk11
+script: clean install
 notifications:
   email:
       - notification@qos.ch
+
+install:
+  - mvn clean install
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ notifications:
   email:
       - notification@qos.ch
 
-  
+script: "mvn clean install"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,6 @@ notifications:
   email:
       - notification@qos.ch
 
-script: "mvn clean install"
+script:
+  - mvn clean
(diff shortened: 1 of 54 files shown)


ffissore:
Uhm build failed in module slf4j-jdk14 but I cannot reproduce

ceki:
FYI, it is better to rebase rather than pull/merge.

ceki:
Instead of pulling in a new dependency, it better more sense to enhance the existing API, more specifically the logger interface

ffissore:
Ok, moved the code into `slf4j-api` module. I've kept module `slf4j-fluent` (renamed to `-test`) because tests depend on logback, and having them in `slf4j-api` would have created a circular dependency.

WRT your suggestion to rebase, I usually do. If you were referring to your commits mixed with mine that is because I based my work on `master` but I'm targeting `2.0`, which is older

ffissore:
For the record, the failed build is again caused by `JDK14MultithreadedInitializationTest`. Before I force-pushed a fix, the build was green, and then it turned red again, so it seems quite erratic

slf4j-api/src/main/java/org/slf4j/Logger.java
```diff
@@ -63,6 +70,23 @@
...
+     *
+     * ...
+     *
+     * logger.atError().log("This is a {}", lazy(() -> "test"));
+     * </pre>
+     *
+     * @param supplier the lambda to cast
+     * @return a Supplier
+     */
+    static Supplier<?> lazy(Supplier<?> supplier) {
```
jliu666: Maybe 
```java
logger.atError().log("This is a {}", () -> "test");
```
is easier to use than 
```
logger.atError().log("This is a {}", lazy(() -> "test"));
```

I think both should be supported.
Why not?

ffissore: It does not compile

jliu666: When will this version release? slf4j community doesn't seem to be active ?

jliu666: add a function to logger:
```java
public void info(String string, Supplier<String> msg);

this compile.
logger.info("This is a {}", () -> "test");
```
More functions in logger makes user less code.

.travis.yml
```diff
@@ -1,7 +1,16 @@
...
 language: java
+
 jdk:
-  - oraclejdk9
+  - oraclejdk11
+
 notifications:
   email:
-      - notification@qos.ch
+      - notification@qos.c
```
twillouer: strange, not ?

ffissore: Good catch. That's 5ddfa9c6b3252834cf3be5f88ea27c821f70a577 which is now on master. @ceki you want to fix that

ceki: @twillouer  thank you. Fixed in the 2.0 branch

PR #206, opened 2019-03-12
Introduces MDC.putAllCloseable

bric3:
This feature offers the same functionality as `MDC.putCloseable` but for
multiple values. And tries to follow the `put`/`putAll` signatures one can
see on a `Map`. This would be extremely useful for structured logging when
the appender is configured to use MDC values.

The `MDCCloseable` class has been modified in such a way that if a single
value is used the consumed memory profile remains the same.

This change is similar to what has been done when `MDC.putCloseable` has been introduced 5c76844e4cd56292f46cabb3f0cfcc39f50253a2. I'm targetting branch 1.7-maintainance, but I'll be happy to port it to other branches e.g. master or 2.0 if that's necessary.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -25,7 +25,9 @@
 package org.slf4j;
 
 import java.io.Closeable;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.slf4j.helpers.NOPMDCAdapter;
 import org.slf4j.helpers.BasicMDCAdapter;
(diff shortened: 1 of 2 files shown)


kr5ture:
Any chances for merging this?

renoth:
Please merge, this is useful.

bric3:
For reference, I moved my code to use [logstash/logstash-logback-encoder](https://github.com/logstash/logstash-logback-encoder) instead as it offers very good json logging in combination with the encoder's [`StructuredArguments` typed API](https://github.com/logstash/logstash-logback-encoder#event-specific-custom-fields) to preserve type in the log event json document (numbers are serialized as numbers not as String).

PR #211, opened 2019-07-29
1.7.26 debug level fix

franciscolopezsancho:
Fix TRACE level in a method `debug`. Not sure, though, how could I ask for a pull request with no conflicts as I created the branch from the tag 1.7.26 already. Hope you can help out @ceki and that it makes sense. Thanks!


ceki:
Fixed in 1.7.27

PR #213, opened 2019-07-30
1.7.x support for JPMS

nhojpatrick:
Can this be merge into the 1.7 release branch and released.

This allows downstream projects to switch to JPMS depending upon an "Automatic-Module-Name", instead of a file based name i.e. slf4j.api.

I've got multiple code branches ready to submit to github projects once this is merged and release.

The only module i really need is the api project, i've guest what you might want other projects to be called and have no problems with them being changed.

### DIFF
--- a/jcl-over-slf4j/pom.xml
+++ b/jcl-over-slf4j/pom.xml
@@ -28,5 +28,8 @@
     </dependency>
   </dependencies>
 
- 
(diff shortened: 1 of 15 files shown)


ceki:
https://jira.qos.ch/browse/SLF4J-420 fixed with a slightly different approach

PR #214, opened 2019-08-08
Logback 1478

degroves:
This change removes the call to initialize the provider in `MarkerFactory.java` as  Ceki Gülcü suggested.  This also adds a new unit tests to make sure the provider is not initialized twice.


ceki:
Hi Dan,

Your proposal is quite good. However, I think the following is less intrusive and does catch the bug.

```java
	@Test
	public void verifyImpactOfMarkerFactory() {
		
		ILoggerFactory firstFactory = LoggerFactory.getILoggerFactory();
		MarkerFactory.getMarker("DOUBLE_INIT");
		ILoggerFactory secondtFactory = LoggerFactory.getILoggerFactory();
		
		if(firstFactory != secondtFactory) {
			fail("MarkerFactory.getMarker causes multiple provider initialization");
		}
	}
```

ceki:
I should also add that my initial solution resembled yours .

If you would like to make further contributions to SLF4J or logback, I propose to send me a scanned copy of a signed ICLA. It can be found at: http://logback.qos.ch/cla.txt

ceki:
Subsequent to our discussion I implemented the changes including the unit test. I apologize for not merging this PR.

PR #215, opened 2019-08-12
Fixes SLF4J-466

delgurth:
- Moved 2 methods from MessageFormatter to Util so we could reuse them

- Fixed issue with one of the EventRecordingLogger warn methods, it was missing the Marker

- Fix SLF4j-466 if last Object passed to the logger is a Throwable it is now placed in the throwable field of the SubstituteLogger. All two object argument log calls had to call a new method in order to do this efficiently.


ceki:
The changes you made in EventRecodingLogger look great. Obvious once you see them, but no so obvious otherwise.

delgurth:
> The changes you made in EventRecodingLogger look great. Obvious once you see them, but no so obvious otherwise.

Thank you. Only thing that was not to my liking is the:

```java
        // 1 and 2 args are covered by other methods
        if (throwable == null && args != null && args.length > 2) {
```
But I didn't see a good way to improve that.

ceki:
Given this contribution, may I ask you to file a [CLA](http://logback.qos.ch/cla.txt)? 

The CLA can be sent to me directly by email. A scanned copy on two separate pages is fine but please make sure that the images render reasonably well when printed on a laser printer. Images taken by cellphone cameras often render as a large gray smudge when printed.

ceki:
Further refactoring removed the tests in  the recordEvent() method altogether. 
See my [recent commit](https://github.com/qos-ch/slf4j/commit/78c10ee651d0c8317971982cbcfdf963aabc8285?diff=unified).

I am sorry for the reformatting making the changes less obvious.

delgurth:
@ceki Thanks, I'll take a look.

Seems something went wrong with the formatting though, seeing the whitespace changes.

delgurth:
I like what you did with it. I should have thought about the switch statement myself for the test case. Saves lot of duplicated code.

PR #223, opened 2019-10-02
Treat a null map as an empty map in MDC.setContextMap

opticyclic:
This is to address https://jira.qos.ch/browse/SLF4J-414 and also https://jira.qos.ch/browse/LOGBACK-944

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -235,6 +235,9 @@ public class MDC {
         if (mdcAdapter == null) {
             throw new IllegalStateException("MDCAdapter cannot be null. See also " + NULL_MDCA_URL);
         }
+        if (contextMap == null) {
+            contextMap = new HashMap<String, String>();
+        }
         mdcAdapter.setContextMap(contextMap);
     }
 


slf4j-api/src/main/java/org/slf4j/MDC.java
```diff
@@ -235,6 +235,9 @@ public static void setContextMap(Map<String, String> contextMap) {
         if (mdcAdapter == null) {
             throw new IllegalStateException("MDCAdapter cannot be null. See also " + NULL_MDCA_URL);
         }
+        if (contextMap == null) {
+            contextMap = new HashMap<String, String>();
```
KengoTODA, line 239: The build in this branch is broken because `HashMap` is not imported in this class.
We need to add `import` to the head of this file, or use FQCN like below:

```suggestion
            contextMap = new java.util.HashMap<String, String>();
```

ceki:
@opticyclic I am assuming your name is Alexei Orischenko. Is that correct?

ceki:
@opticyclic Thank you for this PR. I have opted for a different approach where instead of touching org.slf4j.MDC, implementations of MDCAdapter were modified. See [commit d7786eafa7a5](https://github.com/qos-ch/slf4j/commit/d7786eafa7a5) for details.

PR #226, opened 2019-10-13
this module is Apache 2.0 licensed

hboutemy:
having a pom.xml in sync may help consumers

### DIFF
--- a/jcl-over-slf4j/pom.xml
+++ b/jcl-over-slf4j/pom.xml
@@ -16,6 +16,14 @@
   <description>JCL 1.2 implemented over SLF4J</description>
   <url>http://www.slf4j.org</url>
 
+  <licenses>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
+    </license>
+  </licenses>
+
   <properties>
     <module-name>org.apache.commons.logging</module-name>
   </properties>


ceki:
Thanks Hervé.

eolivelli:
@ceki
Can you cherry pick to 1.7 line?
We need it for Maven binary.
I am also gently asking for a release of 1.7.x

ceki:
I am on it. Perhaps I should mention that deploying on Maven Central is quite slow.

eolivelli:
Thank you @ceki 

"Maven Central is quite slow."
it is due to the provider that is hosting Maven central, as far as I know they are working hard to make the system run well again

ceki:
@eolivelli  Done. 1.7.29 should be available in MavenCentral. By the way, deployoing to MavenCentral  was actually not that bad.

eolivelli:
@ceki 
it looks like 1.7.29 is still not available, did you have problems ?
https://mvnrepository.com/artifact/org.slf4j/jcl-over-slf4j

ceki:
I see it here: http://repo2.maven.org/maven2/org/slf4j/slf4j-api/1.7.29/

eolivelli:
Actually I am able to compile with 1.7.29.
Maybe it is a problem of Maven Search Engine.

Thank you so much!
Have a nice weekend

ceki:
You too.

eolivelli:
@ceki isn't slf4-api Apache License 2.0 as well ? I see it is MIT

https://github.com/qos-ch/slf4j/blob/master/pom.xml

the link to the license is to a php page 
Maybe it is better to make it point to a plain txt file
https://raw.githubusercontent.com/qos-ch/slf4j/master/LICENSE.txt

hboutemy:
> Maybe it is a problem of Maven Search Engine.
not really a problem, but a delay between the artifact availability for download (= what build tool uses) and the misc indexes for search engines

ceki:
slf4j is MIT licensed with the exception of jcl-over-slf4j.

eolivelli:
@ceki see
https://github.com/qos-ch/slf4j/blob/4a1d5194fe63e503959cffb0ba05be0cb26e0eab/pom.xml#L25

ppkarwasz:
> slf4j is MIT licensed with the exception of jcl-over-slf4j.

`log4j-over-slf4j` seems also to be licensed under `Apache-2.0`, is this right?

PR #228, opened 2019-12-10
Fix NOP fallback memory leak

davishmcclurg:
Substitute loggers are not getting "fixed" when no SLF4J provider is
found, which means `SUBST_PROVIDER` will add messages to its
`LinkedBlockingQueue` until the process runs out of memory.

`NoBindingMultithreadedInitializationTest` doesn't contain any
assertions of its own because `MultithreadedInitializationTest` already
asserts loggers are fixed in `assertAllSubstLoggersAreFixed`.

https://jira.qos.ch/browse/SLF4J-469


ceki:
@davishmcclurg Excellent PR. Thank you.

PR #232, opened 2020-03-10
Adapter for Platform Logging (JEP 264)

nipafx:
[JEP 264: Platform Logging API and Service](http://openjdk.java.net/jeps/264) introduced a mechanism to Java 9 that allows third-party logging frameworks to handle JDK log messages (like those Swing produces; not JVM messages). This PR adapts `org.slf4j.LoggerFactory` and `org.slf4j.Logger` to the new interfaces `java.lang.System.LoggerFinder` and `java.lang.System.Logger` and - following [the `ServiceLoader` documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html) - creates a service provider-configuration file and a module declaration, so the integration works from class path and module path.

Closes [SLF4J-442](https://jira.qos.ch/browse/SLF4J-442).

### DIFF
--- a/slf4j-jdk-platform-logging/pom.xml
+++ b/slf4j-jdk-platform-logging/pom.xml
@@ -23,6 +23,19 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>2.0.0-alpha0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>${junit.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/slf4j-jdk-platform-logging/pom.xml
+++ b/slf4j-jdk-platform-logging/pom.xml
@@ -25,4 +25,16 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <!-- target Java 9+ -->
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <release>11</release>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/slf4j-jdk-platform-logging/pom.xml
+++ b/slf4j-jdk-platform-logging/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>slf4j-parent</artifactId>
+    <groupId>org.slf4j</groupId>
+    <version>2.0.0-alpha0</version>
+  </parent>
+
+  <artifactId>slf4j-jdk-platform-logging</artifactId>
+  <packaging>jar</packaging>
+  <name>SLF4J JDK Platform Logging Integration</name>
+  <description>Integrated SLF4J with the Platform Logging API added by JEP 264 in Java 9</description>
+
+  <url>http://www.slf4j.org</url>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/slf4j-jdk-platform-logging/src/main/java/module-info.java
+++ b/slf4j-jdk-platform-logging/src/main/java/module-info.java
@@ -1,0 +1,5 @@
+module org.slf4j.jdk {
+    requires org.slf4j;
+    provides java.lang.System.LoggerFinder
+            with org.slf4j.jdk.SLF4JSystemLoggerFinder;
+}
--- a/slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
+++ b/slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
@@ -3,6 +3,7 @@ package org.slf4j.jdk;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.MissingResourceException;
 import java.util.ResourceBundle;
 
 import static java.util.Objects.requireNonNull;
@@ -53,7 +54,7 @@ class SLF4JSystemLogger implements System.Logger {
 
     @Override
     public void log(Level level, ResourceBundle bundle, String msg, Throwable thrown) {
-        String message = bundle == null ? msg : bundle.getString(msg);
+        String message = getResourceStringOrMessage(bundle, msg);
         switch(level) {
             case ALL:
                 // fall-through intended because a message is visible on all log levels
@@ -84,7 +85,7 @@ class SLF4JSystemLogger implements System.Logger {
 
     @Override
     public void log(Level level, ResourceBundle bundle, String format, Object... params) {
-        String message = bundle == null ? format : bundle.getString(format);
+        String message = getResourceStringOrMessage(bundle, format);
         switch(level) {
             case ALL:
                 // fall-through intended because a message is visible on all log levels
@@ -110,6 +111,25 @@ class SLF4JSystemLogger implements System.Logger {
             default:
                 INTERNAL_LOGGER.error(
                         "SLF4J internal error: unknown log level {} passed to `log` (likely by the JDK).", level);
+        }
+    }
+
+    private static String getResourceStringOrMessage(ResourceBundle bundle, String msg) {
+        if (bundle == null || msg == null)
+            return msg;
+        // ResourceBundle::getString throws:
+        //
+        //  * NullPointerException for null keys
+        //  * ClassCastException if the message is no string
+        //  * MissingResourceException if there is no message for the key
+        //
+        // Handle all of these cases here to avoid log-related exceptions from crashing the JVM.
+        try {
+            return bundle.getString(msg);
+        } catch (MissingResourceException ex) {
+            return msg;
+        } catch (ClassCastException ex) {
+            return bundle.getObject(msg).toString();
         }
     }
 
--- a/slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
+++ b/slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
@@ -1,0 +1,95 @@
+package org.slf4j.jdk;
+
+import org.slf4j.Logger;
+
+import java.text.MessageFormat;
+import java.util.ResourceBundle;
+
+import static java.util.Objects.requireNonNull;
+
+class SLF4JSystemLogger implements System.Logger {
+
+    private final Logger logger;
+
+    public SLF4JSystemLogger(Logger logger) {
+        this.logger = requireNonNull(logger);
+    }
+
+    @Override
+    public String getName() {
+        return logger.getName();
+    }
+
+    @Override
+    public boolean isLoggable(Level level) {
+        switch(level) {
+            case ALL:
+                throw new UnsupportedOperationException();
+            case TRACE:
+                return logger.isTraceEnabled();
+            case DEBUG:
+                return logger.isDebugEnabled();
+            case INFO:
+                return logger.isInfoEnabled();
+            case WARNING:
+                return logger.isWarnEnabled();
+            case ERROR:
+                return logger.isErrorEnabled();
+            case OFF:
+                throw new UnsupportedOperationException();
+        }
+        // TODO
+        return true;
+    }
+
+    @Override
+    public void log(Level level, ResourceBundle bundle, String msg, Throwable thrown) {
+        switch(level) {
+            case ALL:
+                throw new UnsupportedOperationException();
+            case TRACE:
+                logger.trace(msg, thrown);
+                break;
+            case DEBUG:
+                logger.debug(msg, thrown);
+                break;
+            case INFO:
+                logger.info(msg, thrown);
+                break;
+            case WARNING:
+                logger.warn(msg, thrown);
+                break;
+            case ERROR:
+                logger.error(msg, thrown);
+                break;
+            case OFF:
+                throw new UnsupportedOperationException();
+        }
+    }
+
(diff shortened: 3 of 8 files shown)


slf4j-jdk-platform-logging/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>slf4j-parent</artifactId>
+    <groupId>org.slf4j</groupId>
+    <version>2.0.0-alpha0</version>
+  </parent>
+
+  <artifactId>slf4j-jdk-platform-logging</artifactId>
```
nicolaiparlog: I'm not sure what the best name would be for the artifact, module, package, and classes. :blush: At the moment they're a little all over the place using the terms `jdk`, `platform`, and `system`. I think a single term in all places would be better.

slf4j-jdk-platform-logging/src/test/java/org/slf4j/jdk/SLF4JSystemLoggerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        String output = getOutput();
+
+        assertTrue("Captured output should contain logged message.", output.contains(message));
+    }
+
+    private static System.LoggerFinder getSLF4JSystemLoggerFinder() {
+        var loggerFinders = new ArrayList<System.LoggerFinder>();
+        // this fails when test is executed from the module path
+        // because the module declaration does not declare
+        // `uses System.LoggerFinder`
```
nicolaiparlog: At the moment, only the class path integration is tested. Module path integration would fail here.

slf4j-jdk-platform-logging/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,6 @@
+Implementation-Title: slf4j-ext
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: slf4j.ext
+Bundle-Name: slf4j-jdk-platform-logging
+Bundle-Vendor: SLF4J.ORG
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
```
nicolaiparlog: * is that the correct `Bundle-RequiredExecutionEnvironment` or should it be Java 9?
* no `Export-Package`, right?
* oops, forgot `Import-Package`

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
```diff
@@ -0,0 +1,136 @@
...
+                return logger.isWarnEnabled();
+            case ERROR:
+                return logger.isErrorEnabled();
+            case OFF:
+                // all logging is disabled if the highest level is disabled
+                return !logger.isErrorEnabled();
+            default:
+                INTERNAL_LOGGER.error(
+                        "SLF4J internal error: unknown log level {} passed to `isLoggable` (likely by the JDK).", level);
+                return true;
```
nicolaiparlog: I was unsure how to such handle error cases. They're highly unlikely to occur (JDK API should be very stable), but if they do, IMHO SLF4J should not crash, misbehave as little as possible, but let the user know that there's a problem (hence the error logger). See below for two more such cases.

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLoggerFinder.java
```diff
@@ -0,0 +1,34 @@
...
+        //  > is important for platform security. The creator of a logger can
+        //  > pass the class or module for which the logger is created to the
+        //  > LoggerFinder so that the LoggerFinder can figure out which kind
+        //  > of logger to return.
+        //
+        // If backends support the distinction support this distinction and
+        // once `LoggerFactory`'s API is updated to forward a module, we
+        // should do that here.
+        //
+        // [1] https://openjdk.java.net/jeps/264
```
nicolaiparlog: Does this make sense?

slf4j-jdk-platform-logging/src/main/java/module-info.java
```diff
@@ -0,0 +1,5 @@
+module org.slf4j.jdk {
+    requires org.slf4j;
+    provides java.lang.System.LoggerFinder
+            with org.slf4j.jdk.SLF4JSystemLoggerFinder;
+}
```
nicolaiparlog: Should this be in `src/java9`? Would be more consistent but then again, the entire thing only works on Java 9+.

slf4j-jdk-platform-logging/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <!-- target Java 9+ -->
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <release>11</release>
```
nicolaiparlog: Which Java version do you want to target? Java 9 is the most lenient, but I wouldn't encourage use of outdated versions.

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLoggerFinder.java
t1: I'm not sure that I quite get the point: the `System.Logger` API is intended for JDK classes, not for applications, or is it?

The javadoc of `System.LoggerFinder` requires a slightly different thing:

```
     * In addition, when a security manager is present, loggers provided to
     * system classes should not be directly configurable through the logging
     * backend without requiring permissions.
     * <br>
     * It is the responsibility of the provider of
     * the concrete {@code LoggerFinder} implementation to ensure that
     * these loggers are not configured by untrusted code without proper
     * permission checks, as configuration performed on such loggers usually
     * affects all applications in the same Java Runtime.
```

We don't know the logging backends and how they configure things. So we have to check with the `SecurityManager` for _all_ cases, don't we? But this is already checked by the constructor of the `System.LoggerFinder`... I'm confused.

slf4j-jdk-platform-logging/src/test/java/org/slf4j/jdk/SLF4JSystemLoggerTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        System.setErr(ERROR_OUT);
+    }
+
+    @Test
+    public void loggerFinderLoadedAsOnlyService() {
+        // this method asserts that there is exactly one `LoggerFinder` and its of the correct type
```
t1: Is this an actual requirement? It could be over-specified. We only would have to change the `getSLF4JSystemLoggerFinder` method to filter the `SLF4JSystemLoggerFinder` (see below).

slf4j-jdk-platform-logging/src/test/java/org/slf4j/jdk/SLF4JSystemLoggerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        // this fails when test is executed from the module path
+        // because the module declaration does not declare
+        // `uses System.LoggerFinder`
+        ServiceLoader.load(System.LoggerFinder.class).forEach(loggerFinders::add);
+        assertEquals("There should be exactly one `LoggerFinder`.", 1, loggerFinders.size());
+        var loggerFinder = loggerFinders.get(0);
+        assertEquals("The `LoggerFinder` should be of type `SLF4JSystemLoggerFinder`.",
+                     SLF4JSystemLoggerFinder.class,
+                     loggerFinder.getClass());
+        return loggerFinder;
```
t1: Maybe:

```suggestion
        // this fails when test is executed from the module path
        // because the module declaration does not declare
        // `uses System.LoggerFinder`
        return ServiceLoader.load(System.LoggerFinder.class).stream()
            .filter(finder → finder instanceof SLF4JSystemLoggerFinder)
            .findAny()
            .orElseThrow();
```

Note: this `orElseThrow` requires JDK 10.

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLoggerFinder.java
nicolaiparlog: > I'm not sure that I quite get the point: the `System.Logger` API is intended for JDK classes, not for applications, or is it?

It will be used by JDK classes, but it's implemented here and the quote addresses implementations, so I assume it applies to us. Didn't think to look at the JavaDoc, though, and it's not mentioned there, so maybe it's outdated?

Re security manager, it is indeed weird that `System.LoggerFinder` already does the check. Not sure what to do here.

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
```diff
@@ -0,0 +1,136 @@
...
+        }
```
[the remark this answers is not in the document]
nicolaiparlog: The compiler doesn't do that check for switch statements (your IDE may, though). Whether the code is in the `default` branch or not, makes no difference.

slf4j-jdk-platform-logging/src/test/java/org/slf4j/jdk/SLF4JSystemLoggerTest.java
nicolaiparlog: Yeah, that makes sense.

slf4j-jdk-platform-logging/src/main/java/org/slf4j/jdk/SLF4JSystemLogger.java
joachim-durchholz-six: To avoid error cases from future JDK changes, you could compare to `Level.getSeverity()`.
Basic idea is that you don't want to map all j.u.l. levels, you want to start from the available SLF4J levels and determine into which bracket the passed-in j.u.l. level fits. (FINER/FINEST etc. have different severities, despite being mapped to the same j.u.l. level.)

ceki:
Attempt at rebase

ceki:
Rebase attempt

slf4j-jdk-platform-logging/src/test/java/org/slf4j/jdk/SLF4JSystemLoggerTest.java
ceki: I observe the following error:

[ERROR] org.slf4j.jdk.SLF4JSystemLoggerTest.loggerLogsMessage  Time elapsed: 0.003 s  <<< ERROR!
java.util.ServiceConfigurationError: java.lang.System$LoggerFinder: module org.slf4j.jdk does not declare `uses`
        at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:589)
        at java.base/java.util.ServiceLoader.checkCaller(ServiceLoader.java:575)
        at java.base/java.util.ServiceLoader.<init>(ServiceLoader.java:504)
        at java.base/java.util.ServiceLoader.load(ServiceLoader.java:1687)
        at org.slf4j.jdk@2.0.0-alpha4-SNAPSHOT/org.slf4j.jdk.SLF4JSystemLoggerTest.getSLF4JSystemLoggerFinder(SLF4JSystemLoggerTest.java:63)
        at org.slf4j.jdk@2.0.0-alpha4-SNAPSHOT/org.slf4j.jdk.SLF4JSystemLoggerTest.getSLF4JSystemLogger(SLF4JSystemLoggerTest.java:83)
        at org.slf4j.jdk@2.0.0-alpha4-SNAPSHOT/org.slf4j.jdk.SLF4JSystemLoggerTest.loggerLogsMessage(SLF4JSystemLoggerTest.java:49)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:567)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:364)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:272)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:237)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:158)

slf4j-jdk-platform-logging/src/main/java/module-info.java
ceki: java.lang.System.LoggerFinder was introduced in Java9. Thus there would be no point in a multi-release jar.

slf4j-jdk-platform-logging/pom.xml
ceki: I would say 9 as the slf4j project currently requires Java 8 to at runtime and 9 at build time. Requiring 9 for this module seems consistent.

ben-manes:
In the last alpha, using this results in `Illegal configuration-file syntax` error. The `META-INF/services` was changed to an invalid line, `package org.slf4j.jdk.platform.logging.SLF4JSystemLoggerFinder`, whereas it should only be the FQCN per this PR. (I was trying to use this + slf4j-nop for testing)

nipafx:
[Indeed](https://github.com/qos-ch/slf4j/blob/master/slf4j-jdk-platform-logging/src/main/resources/META-INF/services/java.lang.System%24LoggerFinder#L1), that's not gonna work. :)

nipafx:
Created #265.

PR #234, opened 2020-04-12
Update MDC to work with a Map<String, Object>

anastasds:
Related JIRA ticket: [SLF4J-168](https://jira.qos.ch/projects/SLF4J/issues/SLF4J-168)

For cloud development, the ability to log numeric values in JSON logs is essential to be able to do aggregations for statistics on logged numeric values in tools like Kibana. MDC currently internally uses `Map<String, String>`, which prevents that and thus hinders its adoption for this use case.

This PR changes MDC to use `Map<String, Object>` to enable the described use case, for consideration for inclusion in SLF4J 2.x.


anastasds:
Submitted to wrong branch.

PR #236, opened 2020-05-20
Minor change to avoid getLevel to return null

elifazio:
I have got a legacy project that access log4j api directly and does not expect to receive null on getLevel method. So I made a little change on version 1.7.30 to return effective level on getLevel, this avoid nullpointer on the legacy project.


prdoyle:
Please don't do this. If you want the effective level, call `getEffectiveLevel`. My project needs to know whether a level was explicitly set.

ceki:
Thank you for the PR.  Category.getLevel is meant to return the set level for the category/logger. Returning effective level from Category.getLevel would be semantically incorrect. 

More importantly, log4j-over-slf4j is **not** indented as a full replacement for log4j functionality but only as a stop-measure.

PR #245, opened 2020-08-08
SLF4J-414: suggest to add a method returning Optional value

KengoTODA:
Unlike #223, this PR suggests adding a method which returns `Optional` value to solve the [SLF4J-414](https://jira.qos.ch/browse/SLF4J-414).

This suggestion keeps existing documented behavior.
It is also good to return `Optional` value instead of asking users to care for the nullness of the returned value.

After the change, usage of these APIs will be like below:

```java
final Map<String, String> context = MDC.getCopiedContextMap().orElse(HashMap::new);
executor.submit(() -> {
    MDC.setContextMap(context);
});
```

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -228,7 +228,7 @@ public class MDC {
      * parameter must only contain keys and values of type String.
      * 
      * @param contextMap
-     *          must contain only keys and values of type String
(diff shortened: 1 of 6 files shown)


ceki:
@KengoTODA,
Thank you for this PR. As mentioned in PR 223, I have opted for a different approach where instead of touching org.slf4j.MDC, implementations of MDCAdapter were modified catering for null input. See commit d7786eafa7a5 for details. 

As you can imagine, I do not wish to change the signature of the MDC method, even in relation with the return type.

PR #246, opened 2020-08-08
SLF4J-498: Introduce GitHub Actions to run build in the PR

KengoTODA:
We use Travis CI to run the build to verify PRs, but it does not report the result to PR page.
For instance, #232 has a checkmark for each pushed commit, but #236 doesn't have it.
So it is possible that we merge PR which breaks the build without being noticed.

I cannot identify the reason of this problem, so suggest using GitHub Actions as a workaround.
It can integrate the build result to each pushed commit without configuration.

close [SLF4J-498](https://jira.qos.ch/browse/SLF4J-498)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,24 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+      - 1.7-maintenance
+  pull_request:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


KengoTODA:
already resolved by a20a22299da69f4888fb82d0cfa41d15a909ed8e so I'll close this PR.

PR #252, opened 2020-11-19
message supplier

andreika63:
When we run code with 
**logger.debug("heavy message generator")** 
or 
**logger.trace("heavy message generator")**
and debug or trace is not enabled, program have to calculate method argument to throw it away in next step.
Message supplier will speed up code execution ))

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/Logger.java
+++ b/slf4j-api/src/main/java/org/slf4j/Logger.java
@@ -36,6 +36,8 @@ import org.slf4j.spi.DefaultLoggingEventBuilder;
 import org.slf4j.spi.LoggingEventBuilder;
 import org.slf4j.spi.NOPLoggingEventBuilder;
 
+import java.util.function.Supplier;
+
 /**
  * The org.slf4j.Logger interface is the main user entry point of SLF4J API.
  * It is expected that logging takes place through concrete implementations
@@ -120,6 +122,14 @@ public interface Logger {
     public void trace(String msg);
 
     /**
+     * Log a message at the TRACE level.
+     *
+     * @param msgSupplier supplier for the message string to be logged
+     * @since 1.4
+     */
+    public void trace(Supplier<String> msgSupplier);
+
+    /**
      * Log a message at the TRACE level according to the specified format
      * and argument.
      * 
@@ -171,6 +181,16 @@ public interface Logger {
      * @since 1.4
      */
     public void trace(String msg, Throwable t);
+
+    /**
+     * Log an exception (throwable) at the TRACE level with an
+     * accompanying message.
+     *
+     * @param msgSupplier supplier for the message accompanying the exception
+     * @param t   the exception (throwable) to log
+     * @since 1.4
+     */
+    public void trace(Supplier<String> msgSupplier, Throwable t);
 
     /**
      * Similar to {@link #isTraceEnabled()} method except that the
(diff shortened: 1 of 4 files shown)


alekseytomin:
Ops. We are thinking in the same direction.
If I were You I would add `SLF4J-504` into p/r title.

slf4j-api/src/main/java/org/slf4j/Logger.java
```diff
@@ -172,6 +182,16 @@ default public LoggingEventBuilder makeLoggingEventBuilder(Level level) {
...
 
+    /**
+     * Log an exception (throwable) at the TRACE level with an
+     * accompanying message.
+     *
+     * @param msgSupplier supplier for the message accompanying the exception
+     * @param t   the exception (throwable) to log
+     * @since 1.4
+     */
+    public void trace(Supplier<String> msgSupplier, Throwable t);
```
alekseytomin, line 193: Better to write
`public void trace(Throwable t, Supplier<String> msgSupplier);`
In kotlin after this we can write `logger.trace(exception) { "Message" }`

slf4j-api/src/main/java/org/slf4j/Logger.java
```diff
@@ -119,6 +121,14 @@ default public LoggingEventBuilder makeLoggingEventBuilder(Level level) {
      */
     public void trace(String msg);
 
+    /**
+     * Log a message at the TRACE level.
+     *
+     * @param msgSupplier supplier for the message string to be logged
+     * @since 1.4
+     */
+    public void trace(Supplier<String> msgSupplier);
```
alekseytomin, line 130: You have to use 
```
    default void trace(Supplier<String> msgSupplier) {
        if (isTraceEnabled()) {
            trace(msgSupplier.get());
        }
    }
```
We don't know all implementations of `Logger`.

slf4j-api/src/main/java/org/slf4j/helpers/NOPLogger.java
```diff
@@ -25,7 +25,8 @@
 package org.slf4j.helpers;
 
 import org.slf4j.Logger;
-import org.slf4j.helpers.MarkerIgnoringBase;
+
+import java.util.function.Supplier;
```
alekseytomin, line 29: If you use `default` method in Logger - you will have not to implement the methods in NOP

alekseytomin (changes requested):
Will you add info/warn/error ?

slf4j-api/src/main/java/org/slf4j/Logger.java
kbs-empulse: @alekseytomin This would be counter-intuitive for Java programmers though, as in all other cases the throwable is the last parameter.

alekseytomin: Ok. We can write kotlin extensions.

slf4j-api/src/main/java/org/slf4j/Logger.java
andreika63: With default method, method "isTraceEnabled" will called twice.
I thought about it,  but we want more speed

andreika63:
> Will you add info/warn/error ?

"If it OK I will add all other methods" - that's the answer ))))

slf4j-api/src/main/java/org/slf4j/helpers/NOPLogger.java
alekseytomin: Yes. 
We must add the default method (for backward compatibility).
We have to add the method into AbstractLogger (for performance).

slf4j-api/src/main/java/org/slf4j/Logger.java
alekseytomin: Yes.
We must add the default method (for backward compatibility).
We have to add the method into AbstractLogger (for performance).

PR #253, opened 2020-12-02
WIP: SLF4J-504 Add functional logging methods

alekseytomin:
In kotlin we want to use
`logger.trace { "Message with $parameter" } `
instead of
`logger.trace("Message with {}", parameter)`

In this example, I've added `trace` method for this.
If it OK I will add all other methods.


twwwt:
An extension of the API with Supplier-based creation of message texts to log would not only be useful for Kotlin developers, IMHO. Java developers could also take advantage of that.

Tibor17:
This is incredible! Worldwide used library but the project is idle for over one year. Beta and alfa versions have been released for the 1.8.x and 2.0 versions and no more progress. Is it a dead project?

victorherraiz-santander:
This is quite useful and could have and with the default implementation does not break a thing..., I hope this will reach master soon.

ceki:
given that the fluent API accomplishes the same, this change will not be added.

PR #267, opened 2021-09-09
SLF4J-522 - Adds NDC.contains() Method

mwootendev:
Updates the NDC class in slf4j-ext to include a static contains() method
that can check the existing nested diagnostic context for the presence
of a value.

Added unit tests to test the new NDC.contains() method.

### DIFF
--- a/slf4j-ext/src/main/java/org/slf4j/NDC.java
+++ b/slf4j-ext/src/main/java/org/slf4j/NDC.java
@@ -1,7 +1,7 @@
 /**
  * Copyright (c) 2004-2011 QOS.ch
  * All rights reserved.
- *
+ * <p>
  * Permission is hereby granted, free  of charge, to any person obtaining
  * a  copy  of this  software  and  associated  documentation files  (the
  * "Software"), to  deal in  the Software without  restriction, including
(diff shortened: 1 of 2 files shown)


ceki:
@mwootendev Can you kindly describe what type of problems are better solved by NDC instead of MDC?

mwootendev:
@ceki I've actually found the NDC useful for tracing a path through a system. It serves as a good "how did we get here", especially when there are multiple paths that interface with some common services.  For instance, if there is a low-level service for looking up a user, there will probably be multiple higher-level parts of the system that utilize it. If each layer pushes its own NDC context you get a fairly accurate picture of how you got there. For instance, the same logging statement in that user service may end up with different NDCs of [ORDERS USER LOOKUP], [USER REGISTRATION LOOKUP], and [USER ACTIVATION LOOKUP] depending on how that service was reached. This is especially useful for log statements that are not errors and do not have a stack trace.

I tend to differentiate MDC as the "what did we get here", like an order number or user ID. Having the context of "how" can often greatly enhance the value of the "what."

ceki:
@mwootendev OK, I am still trying to understand. So  [ORDERS USER LOOKUP], [USER REGISTRATION LOOKUP], and [USER ACTIVATION LOOKUP] are NDC values? 

If so,  [ORDERS USER LOOKUP] describes the ORDERS -> USER -> LOOKUP path?

mwootendev:
@ceki Yes, that would be correct.

PR #271, opened 2021-12-16
SLF4J-528 Use stackwalker to infer calling class

suchintan:
Opening this as a proof of concept as to a way we could grab the caller class from the caller.

This approach is pretty fast compared to the usual approach of generating a throwable, but it's obviously still slower than just explicitly passing in the class

https://ionutbalosin.com/2018/06/an-even-faster-way-than-stackwalker-api-for-asynchronously-processing-the-stack-frames/

It's also a mega-convenient way of accessing the logger. Even if it's slow, the developer convenience seems quite valuable, and developers should be able to use the fast way if the need arises.

A couple open questions:
1. Is this a good idea? 
2. How to integrate this into the actual library? The current jdk version is set to 8 [in here](https://github.com/qos-ch/slf4j/blob/master/pom.xml#L38). I know we require java 11 for slf4j 2.0 and above, but I don't see where that's enforced in the code

Would love some guidance here. 

JIRA ticket here: https://jira.qos.ch/browse/SLF4J-528

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -387,6 +387,38 @@ public final class LoggerFactory {
         return logger;
     }
 
+    /**
+     * Return a logger named corresponding to the class passed as parameter,
+     * using the statically bound {@link ILoggerFactory} instance.
+     *
+     * <p>
+     * In case the the <code>clazz</code> parameter differs from the name of the
+     * caller as computed internally by SLF4J, a logger name mismatch warning
+     * will be printed but only if the
+     * <code>slf4j.detectLoggerNameMismatch</code> system property is set to
+     * true. By default, this property is not set and no warnings will be
+     * printed even in case of a logger name mismatch.
+     *
+     *
+     * Using Stackwalker to grab the calling class is a convenient and fast way of getting the class information.
+     * https://ionutbalosin.com/2018/06/an-even-faster-way-than-stackwalker-api-for-asynchronously-processing-the-stack-frames/
+     *
+     * This no longer requires a developer to needlessly paste the current class's name into the factory instance.
+     *
+     * @param clazz
+     *            the returned logger will be named after clazz
+     * @return logger
+     *
+     *
+     * @see <a
+     *      href="http://www.slf4j.org/codes.html#loggerNameMismatch">Detected
+     *      logger name mismatch</a>
+     */
+    public static Logger getLogger() {
+        var callerClass = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass();
+        return LoggerFactory.getLogger(callerClass);
+    }
+
     private static boolean nonMatchingClasses(Class<?> clazz, Class<?> autoComputedCallingClass) {
         return !autoComputedCallingClass.isAssignableFrom(clazz);
     }


suchintan:
@ceki would love your thoughts here

hhoffstaette:
I've seen `MethodHandles.lookup().lookupClass()` used for exactly this purpose.

suchintan:
> I've seen `MethodHandles.lookup().lookupClass()` used for exactly this purpose.

Yeah that's one way, I wish we could embed this inside the logger instead so the caller has a clean interface

ceki:
> > I've seen `MethodHandles.lookup().lookupClass()` used for exactly this purpose.
> 
> Yeah that's one way, I wish we could embed this inside the logger instead so the caller has a clean interface

`MethodHandles.lookup().lookupClass()` looks pretty good to me. Anyone are to benchmark the performance?

suchintan:
> > > I've seen `MethodHandles.lookup().lookupClass()` used for exactly this purpose.
> > 
> > 
> > Yeah that's one way, I wish we could embed this inside the logger instead so the caller has a clean interface
> 
> `MethodHandles.lookup().lookupClass()` looks pretty good to me. Anyone care to benchmark the performance?

`MethodHandles.lookup().lookupClass()` returns the current class name, not the calling class name right? ie if we embed this in `LoggerFactory.getLogger()`, it would return `Class<LoggerFactory>`, not `Class<CallingClass>`

I can benchmark stackwalker and this approach and spit out the results here

suchintan:
FYI this approach would work in `kotlin` due to the existance of `inline` functions. If you `inline` `getLogger()`, then `MethodHandles.lookup().lookupClass()` would return the class itself due to inlining.

ceki:
Yes, there is the question of inlining (which existed for a very long time independently of Kotlin) and the question of performance. 

We are trying to help the developer diagnose problems. It follows that we should not add uncertainty where there is already an ongoing search for a bug. 

Nevertheless, this would be a very nice feature to offer. I have been doing some tests on my end  (prior this discussion) but did not have the time to pursue.

ceki:
I am aware of the following potential implementations:

Implementation 1 
performance: about 10 microseconds  per operation
{code}
    // Thread stack trace
   StackTraceElement[] trace = Thread.currentThread().getStackTrace();
   // then step through the StackTraceElement[]
{code}

Implementation 2
performance: about 1 microseconds  per operation
{code}
  // use SecurityManager
    private static final class ClassContextSecurityManager extends SecurityManager {
        protected Class<?>[] getClassContext() {
            return super.getClassContext();
        }
    }
    private static ClassContextSecurityManager getSecurityManager() {
        try {
            return new ClassContextSecurityManager();
        } catch (java.lang.SecurityException sm) {
            return null;
        }
    }

  ClassContextSecurityManager securityManager = getSecurityManager();
  Class<?>[] trace = securityManager.getClassContext();
{code}

Implementation 3
performance: about 2 microseconds  per operation
{code}
    // StackWalker API (new in Java9)
    StackWalker WALKER = StackWalker.getInstance();
    List<StackFrame> frames = WALKER.walk(s -> s.limit(4).collect(Collectors.toList()));
{code}

As far as I know, none of the above are particularly fast.

ceki:
This issue is actually a sore point that needs to be tackled.

suchintan:
I'll reopen it

mvfranz:
> I am aware of the following potential implementations:
> 
> Implementation 1 performance: about 10 microseconds per operation
> 
> ```
>     // Thread stack trace
>    StackTraceElement[] trace = Thread.currentThread().getStackTrace();
>    // then step through the StackTraceElement[]
> ```
> 
> Implementation 2 performance: about 1 microseconds per operation
> 
> ```
>   // use SecurityManager
>     private static final class ClassContextSecurityManager extends SecurityManager {
>         protected Class<?>[] getClassContext() {
>             return super.getClassContext();
>         }
>     }
>     private static ClassContextSecurityManager getSecurityManager() {
>         try {
>             return new ClassContextSecurityManager();
>         } catch (java.lang.SecurityException sm) {
>             return null;
>         }
>     }
> 
>   ClassContextSecurityManager securityManager = getSecurityManager();
>   Class<?>[] trace = securityManager.getClassContext();
> ```
> 
> Implementation 3 performance: about 2 microseconds per operation
> 
> ```
>     // StackWalker API (new in Java9)
>     StackWalker WALKER = StackWalker.getInstance();
>     List<StackFrame> frames = WALKER.walk(s -> s.limit(4).collect(Collectors.toList()));
> ```
> 
> According to the benchmarks I have done, none of the above are below 1 microseconds per operation and rather far from the 100 nanoseconds range I am hoping for.

Isn't method 2 already implemented in `org.slf4j.helpers.Util.getCallingClass()`? Is performance really an issue? If there is a performance issue for a specific use case / class the developer can use a different `LoggerFactory.getLogger()`. The convenience of this new API should not be discounted.

ExE-Boss:
@mvfranz
> Isn't method 2 already implemented in `org.slf4j.helpers.Util.getCallingClass()`? Is performance really an issue?

The issue with method 2 is that it relies on a `SecurityManager`, which is deprecated for removal since **Java 17** as part of [JEP 411](https://openjdk.org/jeps/411).

ceki:
> Isn't method 2 already implemented in `org.slf4j.helpers.Util.getCallingClass()`? Is performance really an issue? 
> If there is a performance issue for a specific use case / class the developer can use a different `LoggerFactory.getLogger()`. 
> The convenience of this new API should not be discounted.

Util.getCallingClass() only gets invoked if the the user sets a system property.

Performance is a significant issue as some users have a `Logger` as an instance field and thus call `LoggetFactory.getLogger` per object instance and not per class instance.

mvfranz:
> Implementation 1 performance: about 10 microseconds per operation
> 
> ```
>     // Thread stack trace
>    StackTraceElement[] trace = Thread.currentThread().getStackTrace();
>    // then step through the StackTraceElement[]
> ```

Since the most likely path is to call `new Exception().getStackTrace()` has any performance runs been performed on this direct call?

mvfranz:
> > Isn't method 2 already implemented in `org.slf4j.helpers.Util.getCallingClass()`? Is performance really an issue?
> > If there is a performance issue for a specific use case / class the developer can use a different `LoggerFactory.getLogger()`.
> > The convenience of this new API should not be discounted.
> 
> Util.getCallingClass() only gets invoked if the the user sets a system property.
> 
> Performance is a significant issue as some users have a `Logger` as an instance field and thus call `LoggetFactory.getLogger` per object instance and not per class instance.

Sorry, I meant those as different thoughts. `org.slf4j.helpers.Util.getCallingClass()` is already implement and could be reused. I did not know that the underlying classes it uses is being removed from the JDK.

Adding `LoggetFactory.getLogger()` as a new API that is slow and not appropriate for class level loggers due to performance does not need to be used in those scenarios.  The developer has alternatives and should make the appropriate decision for their use case.

Having a fast Implementation would be great, I just don't want to miss out on the API due to not finding a perfect solution.

suchintan:
Hey folks, sorry I'm closing this. I don't work in Java too much anymore and I've lost the context. Hopefully someone else can pick it up!

PR #272, opened 2021-12-19
SLF4J-256: Fluent API improvement

adrianshum:
- adding message() in event builder, to allow message to be provided before arguments
- adding arg() as alias of addArgument()
- some typo fixes


ceki:
@adrianshum Thank you for this PR. 

The {{log()}} method with no arguments adds flexibility to the API. However, I am afraid that users will forget to call {{log()}} after populating the event. That is the reasons it was not added in the first place. (I forgot about this reasoning until recently. Sorry about that.)

pedrolamarao:
> @adrianshum Thank you for this PR.
> 
> The `log()` method with no arguments adds flexibility to the API. However, I am afraid that users will forget to call `log()` after populating the event. That is the reasons it was not added in the first place. (I forgot about this reasoning until recently. Sorry about that.)

If static analysis was provided to warn users of incorrect usage, would you consider this to have enough usability?

PR #274, opened 2021-12-23
implement putAll and putAllCloseable

Okeanos:
Implements putAll and putAllCloseable described in [SLF4J-387](https://jira.qos.ch/browse/SLF4J-387) – I ran into the same problem a short while ago that I needed a closeable MDC that discards multiple entries after the try-with resource is closed.

If more tests or other changes are required I'll happily take another look at this.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/MDC.java
+++ b/slf4j-api/src/main/java/org/slf4j/MDC.java
@@ -26,7 +26,9 @@ package org.slf4j;
 
 import java.io.Closeable;
 import java.util.Deque;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.slf4j.helpers.BasicMDCAdapter;
 import org.slf4j.helpers.NOPMDCAdapter;
@@ -74,13 +76,25 @@ public class MDC {
      */
     public static class MDCCloseable implements Closeable {
         private final String key;
+        private final Set<String> keys;
 
         private MDCCloseable(String key) {
             this.key = key;
+            this.keys = null;
+        }
+
+        private MDCCloseable(Set<String> keys) {
+            this.key = null;
+            this.keys = new HashSet<>(keys);
         }
 
         public void close() {
-            MDC.remove(this.key);
+            if (this.key != null) {
+                MDC.remove(this.key);
+            }
+            if (this.keys != null) {
+                this.keys.forEach(MDC::remove);
+            }
         }
     }
 
@@ -124,6 +138,32 @@ public class MDC {
     }
 
     /**
+     * Put a diagnostic context map (the <code>entries</code> parameter) as identified with each
+     * <code>key</code> into the current thread's diagnostic context map. The
+     * <code>entries</code> parameter and its <code>keys</code> cannot be null. The
+     * <code>value</code> of each entry can be null only if the underlying implementation
+     * supports it.
+     *
+     * <p>
+     * This method delegates all work to the MDC of the underlying logging system.
+     *
+     * @param entries entries to put in the map
+     *
+     * @throws IllegalArgumentException
+     *           in case the "entries" parameter is null
+     * @since 2.0.0
+     */
+    public static void putAll(Map<String, String> entries) throws IllegalArgumentException {
+        if (entries == null) {
+            throw new IllegalArgumentException("entries parameter cannot be null");
+        }
+        if (mdcAdapter == null) {
+            throw new IllegalStateException("MDCAdapter cannot be null. See also " + NULL_MDCA_URL);
+        }
+        entries.forEach((k,v) -> mdcAdapter.put(k, v));
+    }
+
+    /**
      * Put a diagnostic context value (the <code>val</code> parameter) as identified with the
      * <code>key</code> parameter into the current thread's diagnostic context map. The
      * <code>key</code> parameter cannot be null. The <code>val</code> parameter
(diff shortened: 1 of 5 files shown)


ceki:
relates to https://jira.qos.ch/browse/SLF4J-387

Okeanos:
@ceki If there's anything from me you need to get this merged please let me know and I'll have a look.

ceki:
It's in the pipeline.

Okeanos:
Updated the pull request to resolve merge conflicts with master and removed an unused import I accidentally introduced. If there's anything else you need to continue please let me know and I'll have a look.

Okeanos:
@ceki is there a chance you or someone else from SLF4J can take a look to get this merged into a 2.x or 2.0.x release?

ceki:
Would you be satisfied with `MDC.putAllCloseable(List<String>)` ?

Okeanos:
I am not sure I follow – do you have a minimal example for:
> It seems to me that the values for the keys might not be known at try-time.

Additionally, what would we use as keys then? The idea/intent is to put known key-value-pairs into the MDC for operations that should have no long-term effect on the thread local MDC but we in the end we can still filter log lines by MDC keys consistently.

ceki:
Here is what I had in mind: 
public class MDCExplore {

```
    public static void main(String[] args) {
        final List keyList = List.of("KEY1", "KEY2");
        MDC.put("x", "y"); // populate before try


        try (Closeable closeable = mdc_closables(keyList)) {
            MDC.put("a", "b");
            MDC.put("KEY1", "val1");
        }
   
      // keys x and a remain populated, KEY1, KEY2 are cleared
    }

    private static Coseable mdc_closables(List keyList) {
        return ....;
    }
```

Okeanos:
I think I understand what you intend, thanks for taking the time! I'll get back to you next week, when I have a little more time to illustrate my own intentions with proposing `putAllCloseable` as natural extension of `putCloseable` and why it would be useful to have around.

Okeanos:
Okay, I had some time to write a real-world inspired example of why I think `MDC.putAllCloseable(Map<String, String> map)` is neat. The example is based on using [Project Reactor](https://projectreactor.io) to make async HTTP calls. The WebClient you see below has its own Thread Pool totally disconnected from the Thread Local and also will reuse connections & threads in said pool.

```java
public String someFunction(final UUID id) {
    ResponseEntity<String> response = webClient
        .get()
        .uri(uriBuilder -> uriBuilder.pathSegment("entry", "{id}").build(id))
        // ... handle request
        .doOnEach(signal -> {
            if (signal.isOnError() ||signal.isOnNext() ) {
                // magic MDCCloseable
                try (var mdc = MDCCloseable.putAllCloseable(signal.getContextView().getOrDefault("MDC", new HashMap<>()))) {
                    var response = signal.get();
                    if (signal.isOnError()) {
                        log.warn(
                            "get pipeline failed {}", "value",
                            kv("response_status", response.getStatusCode()),
                            kv("response_body", response.getBody())
                        );
                    } else {
                        log.debug(
                            "get pipeline did stuff",
                            kv("response_status", response.getStatusCode()),
                            kv("response_body", response.getBody())
                        );
                    }
                }
            } else {
                return;
            }
        })
        //.. do more stuff
        // adds MDC to actual async call stack as context variable
        .contextWrite(Context.of("MDC", Optional.ofNullable(MDC.getCopyOfContextMap()).orElseGet(HashMap::new)))
        .block();
    
    // do stuff with the actual response and inspect it

    return response.getBody();
}
```

If I understood your example correctly, transferring it to my example would look similar to this:

```java
public String someFunction(final UUID id) {
    ResponseEntity<String> response = webClient
        .get()
        .uri(uriBuilder -> uriBuilder.pathSegment("entry", "{id}").build(id))
        // ... handle request
        .doOnEach(signal -> {
            if (signal.isOnError() ||signal.isOnNext() ) {
                var localMdc = signal.getContextView().getOrDefault("MDC", new HashMap<>())

                try (var closeable = MDCCloseable.putAllKeys(localMdc.keySet())) { // insert keys to be removed
                    localMdc.forEach(MDC::put);

                var response = signal.get();
                    if (signal.isOnError()) {
                        log.warn(
                            "something failed {}", "value",
                            kv("response_status", response.getStatusCode()))
                        );
                    } else {
                        log.debug(
                            "some debug",
                            kv("response_status", response.getStatusCode())
                        );
                    }
                }
            } else {
                return;
            }
        })
        //.. do more stuff
        // adds MDC to actual async call stack as context variable
        .contextWrite(Context.of("MDC", Optional.ofNullable(MDC.getCopyOfContextMap()).orElseGet(HashMap::new)))
        .block();
    
    // do stuff with the actual response and inspect it

    return response.getBody();
}
```

It's not a huge difference but there's certainly a couple of "gotchas"/caveats here adding noise to the overall implementation.

I already pointed out that `putCloseable(key, value)` exists and would personally prefer extending the semantics and implementation of that instead of creating a new and fully separate idea.

slf4j-api/src/main/java/org/slf4j/MDC.java
```diff
@@ -156,6 +196,40 @@ public static MDCCloseable putCloseable(String key, String val) throws IllegalAr
...
+     *
+     * @param entries entries to put in the map
+     * @return a <code>Closeable</code> who can remove all originally supplied <code>keys</code>
+     * when <code>close</code> is called.
+     *
+     * @throws IllegalArgumentException
+     *           in case the "entries" parameter is null
+     * @since 2.0.0
+     */
+    public static MDCCloseable putAllCloseable(Map<String, String> entries) throws IllegalArgumentException {
```
ascopes, line 228: Would an override like `java.util.Map.of(String, String, ...)` be useful here?

In many of the places in applications where I have wanted to use this sort of functionality, I have usually wanted to add several entries at once. In this case it would read something like

```java
try (MDC.putAllCloseable(
    "X-Request-ID", request.getHeader("X-Request-ID"),
    "User-Agent", request.getHeader("User-Agent"),
    "From", request.getHeader("From")
)) {
    processRequest();
}
```
versus having to declare this separately.

Okeanos:
> [SLF4J-557](https://jira.qos.ch/browse/SLF4J-557) makes a very valid point.

That is indeed a fair point. I am now wondering how to work around that because the auto-closeable neatness I tried showing in my examples above would be sorely missed :/. A clean, consistent, and predictable API does strike me as more important though.

PR #277, opened 2022-01-24
SLF4J-165 Handling of unknown (custom) priorities is different than i…

DavesMan:
…n log4j (throwing an IllegalStateException than using the effective level of the given Priority / Level / Category)


log4j-over-slf4j/src/main/java/org/apache/log4j/Category.java
```diff
@@ -323,7 +323,8 @@ private int priorityToLevelInt(Priority p) {
         case Priority.FATAL_INT:
             return LocationAwareLogger.ERROR_INT;
         default:
-            throw new IllegalStateException("Unknown Priority " + p);
+            slf4jLogger.warn("Unknown Priority " + p);
+            return LocationAwareLogger.TRACE_INT;
```
DavesMan, line 327: Original JIRA suggests DEBUG level. TRACE will be visible everywhere. So feel free to change the corrected logging level.

PR #279, opened 2022-02-11
Help static analysis tools to detect non-terminated fluent API calls

KengoTODA:
This PR suggests adding internal `@ CheckReturnValue` annotations, and use them to annotate non-terminated fluent APIs.
By these annotations, static analysis tools like [Errorprone](http://errorprone.info/) can detect non-terminated fluent API calls, which should be misuse of SLF4J API.

refs https://github.com/KengoTODA/errorprone-slf4j/issues/143


KengoTODA:
closed by 4e4e56a2df730cdb6877449b6341279208f20a78

KengoTODA:
I found that methods in the `Logger` class still has no annotation, I'll propose another PR to add them as expected.

PR #281, opened 2022-04-01
Fix typo abbreviation for fully-qualified class name.

josephw:
This is an incredibly minor, non-API typo, but it slightly slowed down remembering what it stood for.

Signed-off-by: Joseph Walton <joe@kafsemo.org>


ceki:
Thanks Joseph.

PR #304, opened 2022-09-02
[SLF4J-548] Fix ServiceLoader usage in servlet environment

ppkarwasz:
If both the servlet container and a web application use SLF4J,
`ServiceLoader` calls are susceptible to three problems:

1. The SLF4J copy in the webapp detects the common providers by can not
instantiate them (they implement a different copy of
`SLF4JProviderService`),
2. The SLF4J copy in the common classloader can bind the providers in
the webapp classloader and cause a memory leak,
3. If the server uses a SecurityManager the static initialization of
`LoggerFactory` fails if called by unprivileged code.

This PR should solve these problems.

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -26,20 +26,24 @@ package org.slf4j;
 
 import java.io.IOException;
 import java.net.URL;
+import java.security.AccessControlException;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
 
 import org.slf4j.event.SubstituteLoggingEvent;
 import org.slf4j.helpers.NOP_FallbackServiceProvider;
-import org.slf4j.helpers.SubstituteServiceProvider;
 import org.slf4j.helpers.SubstituteLogger;
-
+import org.slf4j.helpers.SubstituteServiceProvider;
 import org.slf4j.helpers.Util;
 import org.slf4j.spi.SLF4JServiceProvider;
 
@@ -98,11 +102,22 @@ public final class LoggerFactory {
 
     static volatile SLF4JServiceProvider PROVIDER;
 
-    private static List<SLF4JServiceProvider> findServiceProviders() {
-        ServiceLoader<SLF4JServiceProvider> serviceLoader = ServiceLoader.load(SLF4JServiceProvider.class);
+    // Package access for tests
+    static List<SLF4JServiceProvider> findServiceProviders() {
+        final ClassLoader cl = LoggerFactory.class.getClassLoader();
+        final PrivilegedAction<ServiceLoader<SLF4JServiceProvider>> action = () -> ServiceLoader.load(SLF4JServiceProvider.class, cl);
+        final ServiceLoader<SLF4JServiceProvider> serviceLoader = System.getSecurityManager() != null
+                        ? AccessController.doPrivileged(action)
+                        : action.run();
         List<SLF4JServiceProvider> providerList = new ArrayList<>();
-        for (SLF4JServiceProvider provider : serviceLoader) {
-            providerList.add(provider);
+        Iterator<SLF4JServiceProvider> iterator = serviceLoader.iterator();
+        while (iterator.hasNext()) {
+            try {
+                providerList.add(iterator.next());
+            } catch (ServiceConfigurationError | AccessControlException e) {
+                // Short warning
+                Util.report("A SLF4J service provider failed to instantiate:\n" + e.getMessage());
+            }
         }
         return providerList;
     }
--- a/slf4j-api/src/test/java/org/slf4j/LoggerFactoryTest.java
+++ b/slf4j-api/src/test/java/org/slf4j/LoggerFactoryTest.java
@@ -1,0 +1,210 @@
+/*
+ * Permission is hereby granted, free  of charge, to any person obtaining
+ * a  copy  of this  software  and  associated  documentation files  (the
+ * "Software"), to  deal in  the Software without  restriction, including
+ * without limitation  the rights to  use, copy, modify,  merge, publish,
+ * distribute,  sublicense, and/or sell  copies of  the Software,  and to
+ * permit persons to whom the Software  is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The  above  copyright  notice  and  this permission  notice  shall  be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE  SOFTWARE IS  PROVIDED  "AS  IS", WITHOUT  WARRANTY  OF ANY  KIND,
+ * EXPRESS OR  IMPLIED, INCLUDING  BUT NOT LIMITED  TO THE  WARRANTIES OF
+ * MERCHANTABILITY,    FITNESS    FOR    A   PARTICULAR    PURPOSE    AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+ * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE,  ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+package org.slf4j;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Every.everyItem;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.helpers.NOP_FallbackServiceProvider;
+import org.slf4j.spi.SLF4JServiceProvider;
+
+public class LoggerFactoryTest {
+
+    private static String SLF4J_PROVIDER = SLF4JServiceProvider.class.getName();
+    private static String NOP_PROVIDER = NOP_FallbackServiceProvider.class.getName();
+
+    private WebappClassloader child;
+
+    @Before
+    public void setUp() {
+        final URL apiUrl = getLocation(LoggerFactory.class);
+        final URL testUrl = getLocation(LoggerFactoryTest.class);
+        child = new WebappClassloader(getClass().getClassLoader(), new URL[] { apiUrl, testUrl });
+    }
+
+    /**
+     * Tests if SLF4J in a web application is not disturbed by the SLF4J providers of the servlet container.
+     */
+    @Test
+    public void testNotASubtype() throws Exception {
+        final ClassLoader tccl = Thread.currentThread().getContextClassLoader();
+        // both the server and the webapp have `slf4j-api`, but with different providers:
+        // NOP_FallbackServiceProvider only in the server classloader
+        child.delegateToParent(NOP_PROVIDER);
+        // SubstituteServiceProvider in the webapp
+        child.addResource("META-INF/services/org.slf4j.spi.SLF4JServiceProvider", LoggerFactoryTest.class.getResource("substituteServiceProvider.txt"));
+        try {
+            Thread.currentThread().setContextClassLoader(child);
+            final List<?> providers = findServiceProviders(child);
+            final Class<?> slf4jProvider = child.loadClass(SLF4J_PROVIDER);
+            assertThat(providers, everyItem(instanceOf(slf4jProvider)));
+        } finally {
+            Thread.currentThread().setContextClassLoader(tccl);
+        }
+    }
+
+    /**
+     * Tests if SLF4J in a servlet container's common classloader does not use the providers in the webapplications.
+     */
+    @Test
+    public void testMemoryLeak() throws Exception {
+        final ClassLoader tccl = Thread.currentThread().getContextClassLoader();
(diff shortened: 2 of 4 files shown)


ceki:
[SLF4J-568](https://jira.qos.ch/browse/SLF4J-568)

ppkarwasz:
@ceki,
I believe this PR addresses the concerns of the open Jira issue [SLF4J-548](https://jira.qos.ch/browse/SLF4J-548).

ceki:
@ppkarwasz  [SLF4J-568](https://jira.qos.ch/browse/SLF4J-568) is a copy-and-paste of this PR. So yes, this PR and SLF4J-568 are related.

ceki:
@ppkarwasz Indeed, thank you.

boris-unckel:
@ceki Could you check this, please? We have issues to adopt SLF4J 2 in WildFly-Core downstream. The service load based on TCCL is a mess. @ppkarwasz solution solved it for log4j2 API service load of the corresponding impl.
It would be great to get minor release for it.

ceki:
@boris-unckel  Do you have a description of the problem you are facing? Or a reference to the original problem?

ppkarwasz:
@ceki,

The patch also calls `ServiceLoader.load` with a classloader parameter (the classloader that loaded SLF4J), so it doesn't use the TCCL.

It's basically a simplified version of [ServiceLoaderUtil](https://github.com/apache/logging-log4j2/blob/release-2.x/log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java) in `log4j-api`, which is used by other modules, so it has to deal with caller sensitivity problems. `ServiceLoaderUtil` also deals with OSGI directly, whereas SLF4J requires a Service Loader Mediator (Apache Aries) to be present in the OSGI container.

boris-unckel:
Hello @ceki ,

steps to reproduce:

General: setup latest Maven and a Java 11 environment
1) clone and build https://github.com/boris-unckel/slf4j-jboss-logmanager/tree/JBLOGGING-165_upgrade_slf4j2 with 
```mvn clean install```
2) ```git clone https://github.com/wildfly/wildfly-core.git```
2a) optional run to proove everything is OK with WF-Core before:
```mvn clean install -Dsecurity.manager -DallTests -Dmaven.test.failure.ignore=true >../buildlog.txt 2>&1```
3) upgrade pom.xml line 246 (version.org.jboss.slf4j.slf4j-jboss-logmanager) to 1.2.1.Final-SNAPSHOT ( https://github.com/wildfly/wildfly-core/blob/main/pom.xml#L246 )
4) upgrade pom.xml line 256 (version.org.slf4j) to 2.0.2  ( https://github.com/wildfly/wildfly-core/blob/main/pom.xml#L256 )
5) upgrade testsuite/embedded/pom.xml line (version.ch.qos.logback) to 1.4.1 ( https://github.com/wildfly/wildfly-core/blob/main/testsuite/embedded/pom.xml#L41 )
5) ```mvn clean install -Dsecurity.manager -DallTests -Dmaven.test.failure.ignore=true >../buildlog.txt 2>&1```

The issues occur without SecurityManager, too. To check simply remove ```-Dsecurity.manager```.

I thought I could develop a workaround, initializing slf4j early in the log subsystem:
https://github.com/boris-unckel/wildfly-core/tree/upgrade_slf4j2_showcase
but it fails for the embedded server edition and breaks existing usecases. The second downside is the increase of dependencies and length of startup.

Kind regards
Boris

**Comment by ceki** [comment]
@boris-unckel Thanks for the information. I am quite swamped at the moment. I'll deal with this when I can, but it make take a few days.

**Comment by boris-unckel** [comment]
@ceki Hope you're fine. Any updates on this one?

**Comment by ceki** [comment]
> @ceki Hope you're fine. Any updates on this one?

Hi @boris-unckel, I am sorry for not having attended to this PR. It is at the top of my SLF4J todo list. I expect to start working on SLF4J on the week of Monday 14th and there should be a new release before the end of November.

If you wish things to go a little faster, there is also the option of [championing a release](https://github.com/sponsors/qos-ch/sponsorships?tier_id=77436 ).

**Comment by ceki** [comment]
@boris-unckel  Can you please check if release 2.0.4 solves the issue for you? 

Here is a quote from the release notes:

> `LoggerFactory` now uses the class loader that loaded `LoggerFactory` class itself to find providers of type `SLF4JServiceProvider`. Previously, the thread context class loader was used. This was unintentional and the current  behavior more closely matches that of the SLF4J 1.7 series and earlier. This tentatively fixes [SLF4J-544](https://jira.qos.ch/browse/SLF4J-544)  as well as several other related/duplicating issues.

**Comment by boris-unckel** [comment]
Hello @ceki @ppkarwasz 

I have tested it with WildFly-Core and 2.0.4 still has issues. I have ported the ServiceLoaderUtil from Log4j2 and it works immediately. This is the showcase[1]:
https://github.com/boris-unckel/slf4j/tree/serviceloader

* The requirement is to have the API provided by the server, and the implementation by one of the deployed applications.
* The service load must work with *Security Manager activated*, as we still support Java11+
* The server module (API) is known by the deployed application, but the server module(API) does not have a relationship (for example a parent classloader) to the application.

The WildFly-Core integrated version of it is here:
https://github.com/boris-unckel/wildfly-core/tree/upgrade_slf4j2_showcase2

[1] The showcase contains copies of code from log4j2. The commit message shows source URLs.

**Comment by ceki** [comment]
@boris-unckel Can you please clarify whether  slf4j-api version 1.7.x works for  WildFly-Core?

**Comment by ppkarwasz** [comment]
@boris-unckel,

Thanks for the tests. I was pretty sure this patch is equivalent to what we did in Log4j2 (minus the `MethodHandles` stuff that Log4j2 needs because the caller of `ServiceLoaderUtil` and the class itself can be in different JPMS modules). I'll look into it this weekend.

Given the amount of debugging that `ServiceLoaderUtil` has seen (and a couple of bug reports :-) ), I think that including it into SLF4J would be the safest and easiest solution.

Full disclosure: there is still an open Log4j2 bug report (cf. [LOG4J2-3624](https://issues.apache.org/jira/browse/LOG4J2-3624)) that _might_ be related to `ServiceLoaderUtil`, but I was unable to reproduce it.

**Comment by boris-unckel** [comment]
@ceki @ppkarwasz  thanks for your fast reply

> @boris-unckel Can you please clarify whether slf4j-api version 1.7.x works for WildFly-Core?

Yes, sure: It's used in the current production version https://github.com/wildfly/wildfly-core/blob/19.0.0.Final/pom.xml#L257 you can download it here: https://www.wildfly.org/downloads/

It's also a implicit module: https://docs.wildfly.org/27/Developer_Guide.html#how-and-when-is-an-implicit-module-dependency-added

**Comment by ceki** [comment]
@boris-unckel  Slf4j's LoggerFactory in both version 1.7.x and version 2.0.4 (but not 2.0.3) use the same class loader, i.e. the class loader that loaded the  LoggerFactory class to look for the logging implementation. Thus, why would 1.7.36 work in a given context, in this case WildFly, and 2.0.4 fail in the same context?

**Comment by ceki** [comment]
**Update**:  setting pom entries to use slf4j 2.0.4 and logback 1.4.5, the following commands complete successfully.

cd wildcore-core
mvn clean
mvn install -Dmaven.test.skip=true
cd testsuite/embedded/
mvn install

**Comment by boris-unckel** [comment]
@ceki Thanks for checking it. Please don't ```-Dmaven.test.skip=true``` but ```mvn clean install -Dsecurity.manager -DallTests -Dmaven.test.failure.ignore=true >../buildlog.txt 2>&1``` The logging subsystem has tests, there are integration tests and some tests are embedded and CLI client relevant, where logging is needed.
It's late today, I'm posting a gist with the result tomorrow, hopefully morning.

**Comment by ceki** [comment]
@boris-unckel  I have rerun the tests. I am seeing that ` Slf4jLoggingProfilesTestCase.testProfiles` is failing. The strange part is that it looks like `Slf4jServiceActivator` is not being called at all.

**Comment by boris-unckel** [comment]
> @boris-unckel Slf4j's LoggerFactory in both version 1.7.x and version 2.0.4 (but not 2.0.3) use the same class loader, i.e. the class loader that loaded the LoggerFactory class to look for the logging implementation. Thus, why would 1.7.36 work in a given context, in this case WildFly, and 2.0.4 fail in the same context?

@ceki Good morning, including the SecurityManager (why do you run without?) it seems down to last one error. It's not about the classloader, you're correct for this part.

```
* [ERROR] Failures: 
* [ERROR]   Slf4jLoggingProfilesTestCase.testDeploymentConfigurationResource:29->AbstractLoggingProfilesTestCase.testDeploymentConfigurationResource:301->AbstractLoggingProfilesTe* stCase.deploy:321->AbstractLoggingProfilesTestCase.deploy:332->AbstractLoggingTestCase.deploy:210 Failed to deploy logging1.jar: 10 assets: {"WFLYCTL0080: Failed services" => {"* jboss.deployment.unit.\"logging1.jar\".INSTALL" => "WFLYSRV0153: Failed to process phase INSTALL of deployment \"logging1.jar\"
      Caused by: java.util.ServiceConfigurationError: org.jboss.msc.service.ServiceActivator: Provider org.jboss.as.test.integration.logging.profiles.Slf4jServiceActivator could n  ot be instantiated
      Caused by: java.lang.ExceptionInInitializerError
      Caused by: java.lang.IllegalStateException: org.slf4j.LoggerFactory in failed state. Original exception was thrown EARLIER. See also https://www.slf4j.org/codes.html#unsucce  ssfulInit"}}
* [ERROR]   Slf4jLoggingProfilesTestCase.testProfiles:29->AbstractLoggingProfilesTestCase.testProfiles:202->AbstractLoggingProfilesTestCase.deploy:321->AbstractLoggingProfilesTest* Case.deploy:332->AbstractLoggingTestCase.deploy:210 Failed to deploy logging1.jar: 10 assets: {"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"logging1.jar\".INSTALL* " => "WFLYSRV0153: Failed to process phase INSTALL of deployment \"logging1.jar\"
      Caused by: java.util.ServiceConfigurationError: org.jboss.msc.service.ServiceActivator: Provider org.jboss.as.test.integration.logging.profiles.Slf4jServiceActivator could n  ot be instantiated
      Caused by: java.lang.ExceptionInInitializerError
      Caused by: java.lang.IllegalStateException: Unexpected initialization failure
      Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission \"(\"java.io.FilePermission\" \"/home/borisunckel/.m2/repository/org/jboss/s  lf4j/slf4j-jboss-logmanager/1.2.1.Final-SNAPSHOT/slf4j-jboss-logmanager-1.2.1.Final-SNAPSHOT.jar\" \"read\")\" in code source \"(vfs:/content/logging1.jar <no signer certificate  s>)\" of \"ModuleClassLoader for Module \"deployment.logging1.jar\" from Service Module Loader\")"}}
```
and  I assume it fails due to this change, removing the privileged action: https://github.com/qos-ch/slf4j/commit/43a36303e5a2338c22ec9aad5b01a401034eb553#diff-204f23f1c787a5f1f6162f79eae7d8694d723ef7788e23b79ed270a15c677b00L108-L111 

> I am seeing Slf4jLoggingProfilesTestCase.testProfiles failing. The strange part is that it looks like Slf4jServiceActivator is not being called at all.

My guess is ```<version.org.jboss.slf4j.slf4j-jboss-logmanager>1.2.1.Final-SNAPSHOT``` is Maven central/JBoss public rep based, correct? And the following PR is missing, I have tested with a locally built version: https://github.com/jboss-logging/slf4j-jboss-logmanager/pull/16

Some background information regarding the Security Manager, modules and PriviligedAction use: https://docs.wildfly.org/27/WildFly_Elytron_Security.html#Elytron_JSM_Further_Background

**Comment by ceki** [comment]
The version of  slf4j-jboss-logmanager I am using is from https://github.com/boris-unckel/slf4j-jboss-logmanager/ 

The `git log` command yields the following output:
```commit fb6bf609697fd1d28da192c6bd6de581e4a12a19 (HEAD -> JBLOGGING-165_upgrade_slf4j2, origin/JBLOGGING-165_upgrade_slf4j2)
Author: boris-unckel <bu.githubcom@mail.unckel.net>
Date:   Fri Sep 23 17:56:20 2022 +0200

    [JBLOGGING-165] Upgrade slf4j 2.0.4

    Fixes https://issues.redhat.com/browse/JBLOGGING-165
[cut]
```
So it looks like I am using the same version of  slf4j-jboss-logmanager.

boris-unckel:
@ceki I have tested ```mvn clean install -DallTests -Dsecurity.manager -Dmaven.test.failure.ignore=true >../2022-11-23_2131.txt 2>&1``` leading to the error described above: https://github.com/qos-ch/slf4j/pull/304#issuecomment-1325990974

and ```mvn clean install -DallTests -Dmaven.test.failure.ignore=true >../2022-11-24_1647.txt 2>&1``` which runs without any error. Please turn back to the original patch https://github.com/qos-ch/slf4j/pull/304/commits/06d6c81cbcbe4a8d3844cccd09d471f95029a987

ceki:
@boris-unckel  Sorry for not having been clearer earlier. I rather not change the class loader behavior of org.slf4j.LoggerFactory without very solid justification. 

Given Apache (Jakarta) Commnons-Logging experience, SLF4J was created in 2005 chiefly to avoid class loading problems during logging. It does so by adopting the **simplest** class loading model possible. Patch 06d6c81c offers a more dynamic model, and from your previous comments, which is apparently not even needed. 

As for using SecurityManager that is a different question to which I am open.

ppkarwasz:
> @boris-unckel Sorry for not having been clearer earlier. I rather not change the class loading behavior of `org.slf4j.LoggerFactory` without very solid justification.

@ceki, the following code does not change the class loading behavior: since the `action` lambda was created by `LoggerFactory`, running `action` is equivalent to calling the function directly. The `SecurityManager` part is required by WildFly, because the code that initiates SLF4J is not privileged enough to access Logback, while `slf4j-api` is.

```lang-java
final ClassLoader cl = LoggerFactory.class.getClassLoader();
final PrivilegedAction<ServiceLoader<SLF4JServiceProvider>> action = () -> ServiceLoader.load(SLF4JServiceProvider.class, cl);
final ServiceLoader<SLF4JServiceProvider> serviceLoader = System.getSecurityManager() != null
    ? AccessController.doPrivileged(action)
     : action.run();
```

I am far from being an expert on `SecurityManager`, but I believe that Boris known what he is doing.

boris-unckel:
@ceki Yes, sure. It was meant only regarding the Security Manager, the ClassLoading works as the tests have shown. 
@ppkarwasz thanks for the snipplet

ceki:
Cool. We are all in agreement then.

boris-unckel:
@ceki Just to be sure: You're preparing the fix? Or do you expect a PR?

ceki:
I am preparing a fix.

boris-unckel:
Thanks to you both for all your support.

ceki:
@boris-unckel  Can you please verify that 3bc58f3e81cf committed a few moments ago works?

boris-unckel:
@ceki Thanks, I have prepared a local snapshot build and the test is running right now. I'm going to give a status with the result tomorrow morning.

boris-unckel:
@ceki The test run completed successfully. Thanks for your ongoing support to get this solved.

prrvchr:
Thanks for making the SLF4J API embeddable in a jar archive...

PR #305, opened 2022-09-09
Help static analysis tools to detect non-terminated fluent API calls

KengoTODA:
This PR suggests using internal `@ CheckReturnValue` annotations to annotate non-terminated fluent APIs.
By these annotations, static analysis tools like [Errorprone](http://errorprone.info/) can detect non-terminated fluent API calls, which should be misuse of SLF4J API.

refs https://github.com/KengoTODA/errorprone-slf4j/issues/143


ceki:
PR without a Jira issue have a tendency to fall in between the cracks.

PR #310, opened 2022-09-25
Add missing javadoc to SLF4JServiceProvider.java

ascopes:
Had to do some digging around the logback-classic alpha release for SLF4J2 API integration to work out the nature of this behaviour. This should prevent confusion for developers of other logging implementations in the future.


ascopes:
Assuming I shouldn't be the one to mark that sign-off on the commit?

If I need to do anything else, please let me know :)

ceki:
> Assuming I shouldn't be the one to mark that sign-off on the commit?

In fact, the sign-off is a [Developer Certificate of Origin](https://developercertificate.org/). So, the sign-off must be performed by the contributor.

ascopes:
@ceki got it! Will sort that out now, thanks

PR #314, opened 2022-10-06
Update maven plugins

jorsol:
Update some maven plugins and also include the maven-enforcer-plugin to require Java 9+ at build time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,17 +50,17 @@
     <reload4j.version>1.2.22</reload4j.version>    
     <log4j.version>1.2.17</log4j.version>
     <logback.version>1.2.10</logback.version>
-    <junit.version>4.13.1</junit.version>
-    <maven-site-plugin.version>3.7.1</maven-site-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -164,12 +164,12 @@
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>animal-sniffer-maven-plugin</artifactId>
-          <version>1.14</version>
+          <version>1.22</version>
           <configuration>
             <signature>
               <groupId>org.codehaus.mojo.signature</groupId>
               <artifactId>java16</artifactId>
```
slawekjaranowski, line 171: Should be `java18`

By the way plugin looks like not used - only defined in `pluginManagement` 
no `executions`, no reference in `build/plugins`

jorsol: Yes, looks like it's not used, and according to the project:
> this plugin is now in maintenance level as the exact same feature can be now easily achieved using the --release flag from Javac

Removing the declaration...

PR #320, opened 2022-10-28
enhance quality of Manifest

stbischof:
use annotations to define:
-Exports (ManifestExports and Imports)
-ServiceProvider (serviceLoader files, Manifest Capabilities) -ServiceConsume (Manifest Requiroments)

This is the first Step.

If you are interested in this way I will finish this an do the cleanup and osgi integration tests.
will compare all old new entry and the changes of the jar.
It might also be discussed if the export of the implementation packages is really needed - since everything is accesablie over the Serviceloader


ceki:
Hi @stbischof 

Thank you for this PR. I have not studied the actual implications but I would like to avoid adding new dependencies to slf4j-api even if these dependencies are annotations.

stbischof:
This are just build time dependencies. Will not have/require them in runtime.

stbischof:
so feel free to close.
i opened https://github.com/qos-ch/slf4j/pull/321 that does the really necessary things manually

PR #327, opened 2022-12-14
Fully automate OSGi metadata creation and implement minor fixes

HannesWell:
This is a follow up PR to https://github.com/qos-ch/slf4j/pull/324 to only use the `maven-bundle-plugin` to generate all OSGi metadata into slf4j's Manifest.MF files. This unifies the resulting MANIFEST.MFs that are currently partly generated and partly statically defined and fixes the following aspects:

- Restore the `Bundle-SymbolicName` from SLF4J-2.0.5 and before
- Generate the OSGi 'Service Loader Mediator' capabilities for all SLF4JServiceProviders using BND-Tools
- Removes Exported-Packages: META-INF.versions.9
- Restricts the slf4j self import versions to [<current-version>,3)
- Add Automatic-Module headers for those artifacts that don't have a module-info.java
- Replace the deprecated 'Bundle-RequiredExecutionEnvironment' by a corresponding required 'osgi.ee' capability.
- Fix the package-export of slf4j-jdk-platform-logging (from 'slf4j.jdk.platform.logging' to 'org.slf4j.jdk.platform.logging')
- Adds the 'Main-Class' header for the migrator artifact again(?)

Some of the fixed issues were introduced with https://github.com/qos-ch/slf4j/pull/324 (mainly the changed BSN), some existed before.

I'm currently unsure if it is sufficient to have Import-Package version ranges for the slf4j packages that only contain the minor version (like `version="[2.0,3)"`) or if it should be better also include the current service/micro version in the lower bound (like `version="[2.0.7,3)"`. That latter one is what this PR is currently doing, but the former would be simpler because it is what BND-tools (which the maven-bundle-plugin uses under the hood) is doing by default. However my first impuls would be that a library should never (self-)import a lower version of itself.

@stbischof and @timothyjward you made some OSGi related enhancements of SLF4J in the recent past and are maybe also interested in this change.

@ceki do I have to create a extra ticket for this?

PR #328, opened 2022-12-14
Format pom.xml files and remove trailing whitespace

HannesWell:
This is the part of https://github.com/qos-ch/slf4j/pull/327 that only applies some white-space clean ups and does one re-ordering of elements. Just in case it should be reviewed separately.


ceki:
@HannesWell  Hi, I would not worry about a few space changes. Not merging.

HannesWell:
OK. Thanks for the clearification.
Closing.

PR #334, opened 2023-01-12
Add 'Main-Class' header to slf4j-migrator Manifest.MF

HannesWell:
Because the Manifest.MF did not have a terminating new-line it was invalid and its content was ignored.

Disable the execution of the 'maven-bundle-plugin' for the migrator module to not have (useless) OSGi-metadata generated into it.

### DIFF
--- a/slf4j-migrator/pom.xml
+++ b/slf4j-migrator/pom.xml
@@ -16,4 +16,20 @@
   <name>SLF4J Migrator</name>
   <description>SLF4J Migrator</description>
 
+  <build>
+    <plugins>
+
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


HannesWell:
@ceki maybe do you want to have a look at this as well. Because of this the migration can probably not used as described here:
https://www.slf4j.org/migrator.html

Alternatively this entry could be generated using the maven-jar-plugin, but since the intention is to use the maven-bundle-plugin in all other plugins to generate Manifest-entries (#340) I choose to use it here as well.
```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-jar-plugin</artifactId>
        <configuration>
          <archive>
            <manifest>
              <mainClass>org.slf4j.migrator.Main</mainClass>
            </manifest>
          </archive>
        </configuration>
      </plugin>
```

PR #336, opened 2023-01-31

m-froehlich:
The LoggerFactory will load service providers from a passed in ClassLoader, if given. Otherwise previous behavior is maintained. The getLogger(Class) method receives a class, that is used to extract a ClassLoader from.

Signed-off-by: Marvin Fröhlich m.froehlich@infolog.de

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
+++ b/slf4j-api/src/main/java/org/slf4j/LoggerFactory.java
@@ -104,16 +104,45 @@ public final class LoggerFactory {
     static volatile SLF4JServiceProvider PROVIDER;
 
     // Package access for tests
-    static List<SLF4JServiceProvider> findServiceProviders() {
+    static List<SLF4JServiceProvider> findServiceProviders(ClassLoader classLoader) {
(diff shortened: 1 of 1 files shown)


m-froehlich:
I have created another patch branch with the same changes, but with Singed-off-by line in the commit message.

What checks haved failed? What do I need to do?

ceki:
This change introduces behavior which is subtly different than older code. It will not be merged.

m-froehlich:
Without my patch slf4j simply doesn't work in combination with a custom class loader.

m-froehlich:
Can you please comment on this? This really is a bug in the code, that needs to be fixed.

Do I have to do anything to make this fix good to be merged?

ceki:
Your patch changes the public API. Moreover, it modifies existing behavior and has a much bigger impact than you seem to think. 

It will not be merged. I can't be clearer than this.

m-froehlich:
Are you open to an API extension on that front, that can be used optionally to workaround this bug? Hopefully the guys at PAC4J will make use of it then.

PR #324, opened 2022-12-09
SLF4J-574: Add full OSGi headers, especially "uses" clauses

jonahgraham:
Fixes: https://jira.qos.ch/browse/SLF4J-574

This uses the BND tool, via org.apache.felix to generate the MANIFEST.MF to be fully compliant with OSGi.
Compared to using just the maven-jar-plugin to create the MANIFEST.MF the uses clauses are added to the
Export-Package

PR #340, opened 2023-03-18
Fully automate OSGi metadata creation and restore OSGi-metadata in remaining slf4j modules

HannesWell:
This is a follow-up of https://github.com/qos-ch/slf4j/pull/324 and the reduced version of https://github.com/qos-ch/slf4j/pull/327, that has not been covered in https://github.com/qos-ch/slf4j/pull/330.

With this change the OSGi metadata of all remaining slf4j modules are generated using the maven-bundle-plugin.
This makes all remaining MANIFEST.MF files obsolete and they are therefore removed. Then sfl4j only has one source for OSGi metadata: the configuration of the maven-bundle-plugin. This simplifies the understanding how the metadata are crafted and maintenance on the long run, as requested in https://github.com/qos-ch/slf4j/pull/331#issuecomment-1472395543 and noted in [SLF4J-583](https://jira.qos.ch/browse/SLF4J-583).

This also re-adds metadata that where accidentally removed in 2.0.6 and thus fixes [SLF4J-578](https://jira.qos.ch/browse/SLF4J-578).
Additionally it improves the metadata by adding more use-clauses to more Export-Package entries.

@ceki can you please review this or let me know if you want to have this split up? I know you prefer smaller steps, but I didn't wanted to create one PR for each slf4j module and therefore made a bigger change. Nevertheless the single changes are now simpler compared to other PRs.
The PR is build in two steps. The first commit is dedicated to restore missing metadata and to remove entries from the manifest that are already generated by the m-bundle-plugin and are therefore duplicates. This already allowed to remove some Manifests entirely.
The second commit then replaces all remaining manifests by a corresponding configuration of the maven-bundle-plugin, so that no OSGi metadata are crafted from that. In order to simplify the generation of the OSGi Service Loader Mediator headers I created a profile that is activated by the existence of the file `src/main/resources/META-INF/services/org.slf4j.spi.SLF4JServiceProvider` and adds the corresponding manifest headers. The properties in the header are filled with values from the corresponding project.

Please let me know if you want a change or something is not clear.

### DIFF
--- a/jcl-over-slf4j/pom.xml
+++ b/jcl-over-slf4j/pom.xml
@@ -27,7 +27,7 @@
   <properties>
     <module-name>org.apache.commons.logging</module-name>
   </properties>
-  
+
   <dependencies>
     <dependency>
       <groupId>org.slf4j</groupId>
@@ -39,5 +39,19 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
 
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <_exportcontents>org.apache.commons.logging*;version=${jcl.version};-noimport:=true</_exportcontents>
+          </instructions>
+        </configuration>
+      </plugin>
+
+    </plugins>
+  </build>
 </project>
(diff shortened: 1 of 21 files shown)


HannesWell:
Thank you Ceki for your review and merging this.
I checked the generated OSGi metadata again and they look much better now. 👍🏽

PR #273, opened 2021-12-22
Platform Logging bridge should use MessageFormat but it's using String.format

miyakawataku:
[java.lang.System.Logger.log](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.Logger.html#log(java.lang.System.Logger.Level,java.lang.String,java.lang.Object...)) is specified to accept the message format of MessageFormat. However, the current SLF4JPlatformLogger implementation uses String.format instead of MessageFormat. As a result, the current implementation cannot format the message correctly.

# Expected  and actual behavior

```java
LoggerFinder finder = System.LoggerFinder.getLoggerFinder();
Logger systemLogger = finder.getLogger("smoke", null);
systemLogger.log(Level.INFO, "hello {0}", "world");
```

Expected output:

```
INFO smoke - hello world
```

Actual output:

```
INFO smoke - hello {0}
```

# What this change does

This change replaces use of String.format by MessageFormat.format to comply the spec of Platform Logging.

PR #351, opened 2023-07-16
Fix text formatting in slf4j-jdk-platform-logging to use the correct JDK convention

Mechite:
While there is a comment here showing that it was understood the JDK has a different convention, the wrong format method was used here. The JDK uses MessageFormat.

I have not tested this PR and cannot assure that I have made the correct change, a reviewer should test this fork and see if invoking the system logger with extra parameters, i.e. `log("Hello, {0}!", "world")` outputs correctly.


Mechite:
Alright, so after further review of my own PR
- I didn't sign this off with DCO as I missed out the contribution instructions in the README - if this is denied I'll go and make a Jira ticket for it etc
- The test suite will likely fail because it is using the wrong formatting convention as well!

Mechite:
This is a duplicate of https://github.com/qos-ch/slf4j/pull/273, so it is being closed. The diff of the other one is completely identical.

PR #385, opened 2024-01-08
Use computeIfAbsent with a protected method, make reset protected

gnodet:


gnodet:
Thx @ceki for merging this quickly.
Maven should be able to switch to next 2.0.x release now !

Issue #390, opened 2024-01-25
Message formatting with last argument being Throwable


zorba128:
I really often hit issue with slf4j interpreting my messages not the way I'd like it to.
I don't think its reasonable to have two following lines result in different behaviour.
```
Throwable ex = ...
Item item = ...
log.info("Problem {} processing {}", ex, item);
log.warn("Problem processing {}: {}", item, ex);
```

I think the `MessageFormatter` should take into account number of placeholders, and try last to use last argument as throwable candidate only after all of them were already used.

[https://www.slf4j.org/faq.html#paramException] says:

> If the exception is not the last argument, it will be treated as a plain object and its stack trace will NOT be printed. However, such situations should not occur in practice. 

For me this is not true. I quite often want to see just `ex.toString`, specially where exception is part of business logic. I just want to log its name, description, details - I don't care about stack trace. So it's common to treat exception just as any other part of message.

> SLF4J, in accordance with the programmer's most probable intention, will interpret NumberFormatException instance as a throwable instead of an unused Object parameter

And this one actually suggests what should be done. It says about choice between exception and unused object parameter. While code actually does not check if parameter is used - it just picks last argument, resulting in message with last '{}' not filled in, and with unwanted stacktrace appended.

BrandaoFelipe:
Hi,
I’ve been reading through this issue and noticed there were previous attempts in forks
(e.g. zorba128) to improve the handling of Throwable arguments in MessageFormatter.

Before going deeper, I’d like to ask:
Is this issue still considered open by the maintainers?
Would a PR proposing a change in MessageFormatter (with tests) still be welcome?

Either way, I’m planning to experiment locally to better understand the problem.

Thanks!

ceki:
@zorba128 @BrandaoFelipe 

Thank you for this issue and the various pull requests. 

I am not keen to change the code. The present logic assumes that if the exception is the last argument of a printing method call, i.e. trace(), debug() info(), etc..,  then it will be printed as an exception with a stack trace. It is straightforward to explain and to understand. 

If you wish to output the exception message without the stack trace you could write:

`log.warn("Processing issue; item={}, ex={}{}", item, ex, ""); `

or alternatively using the fluent API:

`log.atWarn().addArgument(item).addArgument(ex).log("Processing issue; item={}, ex={}");`

and if you wish to print the stack trace

`log.atWarn().addArgument(item).setThrowable(ex).log("Processing issue; item={}");`

ceki:
[...]

PR #391, opened 2024-01-27
Improved handling of throwables in message formatting.

zorba128:
[Related issue](https://github.com/qos-ch/slf4j/issues/390)

This PR proposes improved exception extraction during message formatting.
It simplifies whole process and improves performance (avoids unnecessary construction of NormalizedParameters, avoids copying/truncating argsArray).

Still some work to be done - logger implementations can now be simplified in many cases.


zorba128:
relates to #166, but I believe this PR solves issue in more straightforward and efficient way (no need to double-parse message).

PR #393, opened 2024-02-04
fix: Explicit provider fails when used with module system

### LINKED ISSUE #392: Explicit provider fails when used with module system
it looks like it is trying to create e new instance based on this property. I guess it would be better to loop over the providers and check if there is a matching instance.

```
SLF4J(I): Attempting to load provider "org.slf4j.simple.SimpleServiceProvider" specified via "slf4j.provider" system property
SLF4J(E): Failed to instantiate the specified SLF4JServiceProvider (org.slf4j.simple.SimpleServiceProvider)
SLF4J(E): Reported exception:
java.lang.IllegalAccessException: class org.slf4j.LoggerFactory (in module org.slf4j) cannot access class org.slf4j.simple.SimpleServiceProvider (in module org.slf4j.simple) because module org.slf4j.simple does not export org.slf4j.simple to module org.slf4j
        at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:394)
        at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:714)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:495)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
        at org.slf4j@2.0.10/org.slf4j.LoggerFactory.loadExplicitlySpecified(LoggerFactory.java:225)
```

rfscholte:
this fixes #392

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-parent</artifactId>
-    <version>2.0.11-SNAPSHOT</version>
+    <version>2.0.11</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-parent</artifactId>
-    <version>2.0.10</version>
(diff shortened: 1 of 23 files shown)


ceki:
Going directly to the explicitly specified provider skips the initialization of available providers, saving a few hundred milliseconds at initialization time.

PR #394, opened 2024-02-05
Pass key-value properties into events in slf4j-reload4j.

josephw:
I'm aiming to migrate an existing codebase off Log4j as a logging framework. Logging is through SLF4J, but there's also extensive direct use of Log4j's `org.apache.log4j.MDC` to pass values (other than `String`s) through to a `Layout` that produces structured logging output.

An incremental migration would be significantly easier if I were able to adopt SLF4J's new fluent API without having to simultaneously change the backend logging framework.

This PR makes exactly that change: the key-value properties from the fluent API are merged into the snapshot of the MDC, leaving the existing logging backend with exactly the same data, while the logging calls lose their direct dependency on Log4j.

It's not clear to me whether this approach was considered and rejected, because it conflates the two different types of property, or whether it's suitable as a general purpose change.

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-parent</artifactId>
-    <version>2.0.11-SNAPSHOT</version>
+    <version>2.0.11</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-parent</artifactId>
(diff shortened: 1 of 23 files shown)


josephw:
This approach has worked well: I've incrementally migrated the codebase away from direct use of Log4j 1.x to pure SLF4J 2.x, without losing any existing structured logging, or needing any whole-codebase changes. It's now possible to move to another logging framework without changing the application code that's producing logs.

This is currently implemented with a fork of `slf4j-reload4j`, which could be dropped if this were applied, but that can go away as soon as the logging framework changes to something currently supported.

josephw:
@ceki any thoughts on merging this? We've now moved to an all-SLF4J codebase, with key-value pairs for all structured logging, but not having this change upstream means we're on a fork of `slf4j-reload4j`. At some point, we'll likely migrate to Logback and drop this entirely, but it'd be good to lose the fork in the meantime.

PR #396, opened 2024-02-12
docs: add docs for error with arguments and throwable

liran2000:
It is common action to log an error message with arguments, together with the exception.  
With current [error method](https://github.com/qos-ch/slf4j/blob/master/slf4j-api/src/main/java/org/slf4j/Logger.java#L793) declaration and _javadoc_, it is hard to tell how to do this.  
This PR adds documentation to the method for explaining it.

Example usage:
```
String arg1 = "v1";
String arg2 = "v2";
log.error("arg1: {}, arg2: {}", arg1, arg2, new Exception("error message"));
```

### DIFF
--- a/slf4j-api/src/main/java/org/slf4j/Logger.java
+++ b/slf4j-api/src/main/java/org/slf4j/Logger.java
@@ -785,7 +785,25 @@ public interface Logger {
      * (and relatively small) cost of creating an <code>Object[]</code> before invoking the method,
      * even if this logger is disabled for ERROR. The variants taking
      * {@link #error(String, Object) one} and {@link #error(String, Object, Object) two}
-     * arguments exist solely in order to avoid this hidden cost.
+     * arguments exist solely in order to avoid this hidden cost. <br/>
+     * <br/>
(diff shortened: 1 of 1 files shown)


liran2000:
hi @ceki would you be able to review this ?

ceki:
@liran2000 Thank you for this PR. Have you seen the comments at the top of the Logger class which mention handling of exceptions? Also the handling of a Throwable passed as last argument is expected to be the same for all printing methods, not just error(..,)

liran2000:
> @liran2000 Thank you for this PR. Have you seen the comments at the top of the Logger class which mention handling of exceptions? Also the handling of a Throwable passed as last argument is expected to be the same for all printing methods, not just error(..,)

I see now. if references to the FAQ, where it is written, without a [full-link](https://www.slf4j.org/faq.html#paramException). For developers at IDE, commonly looking at the method, it can be less clear, and even if developer had the chance to look at the  comments at the top of the Logger class, he then need to search for a link and follow it.

thanks for your response.

PR #405, opened 2024-03-15
Support nested putCloseable invocations (issues/404)

### LINKED ISSUE #404: putCloseable should not remove the key from MDC if it was previously set to a different value
Our team wrote some code that broadly looked like this:

```
MDC.put("color", "red");

try(ignored = MDC.putCloseable("color", "green")) {
...
}

assertTrue(MDC.get("color", "green")) // fails, the key is deleted
```

It should be an easy patch to allow callers to nest putCloseable calls and get expected results; in fact, I'll go try to write one now.

Thank you very much, by the way, for SLF4j -- it's a wonderful tool. Cheers

mdpearson:
Hi folks,

This PR fixes #404 by preventing `putCloseable` from clobbering a key that was present in the MDC before the `putCloseable` was created. It thereby allows nested `putCloseable`s. The existing behavior surprised us when we were writing some recursive code. Let me know what you think, and thanks for making SLF4j!

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.slf4j</groupId>
     <artifactId>slf4j-parent</artifactId>
-    <version>2.0.12-SNAPSHOT</version>
+    <version>2.0.12</version>
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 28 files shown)


rzabson:
Hi @mdpearson!
Is this PR still valid? I see that is no conflicts and this PR could be merged easily.

Issue #425, opened 2024-08-14
slf4j-jdk14 adapter: Logger.log's fqcn parameter ignored


thomas-k-git:
Dear all,
In the process of upgrading from slf4j1 to v2 of api and impl jars, we notice a difference in behavior, breaking our setup.

we have a complex setup with multiple adapters: jboss-logging into log4j2-to-slf4j into slf4j-jdk.

indirectly, org.slf4j.spi.LocationAwareLogger#log is called with the outmost logger class as fcqn parameter.

from my understanding the intention of the parameter is to indicate the boundary / entrypoint into the log framework and the next stackframe past this fcqn should be used as "calling class".

however in our setup, a middle layer of the log adapters is being set as java.util.logging.LogRecord#setSourceClassName and not the actual caller.

so with a callstack setup like this:
```
steArray = {StackTraceElement[131]@3780} 
 0 = {StackTraceElement@3781} "org.slf4j.jul.JDK14LoggerAdapter.fillCallerData(JDK14LoggerAdapter.java:182)"
 1 = {StackTraceElement@3782} "org.slf4j.jul.JDK14LoggerAdapter.innerNormalizedLoggingCallHandler(JDK14LoggerAdapter.java:155)"
 2 = {StackTraceElement@3783} "org.slf4j.jul.JDK14LoggerAdapter.log(JDK14LoggerAdapter.java:172)"
 3 = {StackTraceElement@3784} "org.apache.logging.slf4j.SLF4JLogger.logMessage(SLF4JLogger.java:245)"
 4 = {StackTraceElement@3785} "org.jboss.logging.Log4j2Logger.doLog(Log4j2Logger.java:54)"
 5 = {StackTraceElement@3786} "org.jboss.logging.Logger.infov(Logger.java:1034)"
 6 = {StackTraceElement@3787} "ACTUALCLASS"
```

we get
2024-08-14 14:22:14 INFO  [org.apache.logging.slf4j.SLF4JLogger] message
instead of
2024-08-14 14:22:14 INFO  [ACTUALCLASS] message

even though org.jboss.logging.Logger is the fcqn calling class.

it seems that behavior changed between version 1 and 2 [here](https://github.com/qos-ch/slf4j/commit/9928d80daf022e6461647109d62e237498f0270c?diff=split&w=0#diff-aba54a51d8acd33e2034b9c44686ec64954985d032d3998897f1f285139c6ab7R605)

[...]

PR #438, opened 2024-10-23
Makes `SubstituteLogger` location aware

ppkarwasz:
Currently, `SubstituteLogger` neither supports location information from upstream callers nor passes location information to the delegated logger.

This PR:

- implements `LocationAwareLogger`, which allows upstream callers to pass location information both through the `LocationAwareLogger` API and the fluent API.
- sets the correct caller boundary, when the classic API methods are invoked.

This is related to #425.


ceki:
@ppkarwasz Thank you for this PR. 

For the moment, I fail to see the circumstances under which the LocationAwareLogger.log method may be called. Swill investigating though.

ppkarwasz:
> For the moment, I fail to see the circumstances under which the `LocationAwareLogger.log()` method may be called. Still investigating though.

I am not entirely sure when `SubstituteLogger` is used (or if it is used at all these days), but I assume that it could appear in OSGi environments if:

1. The SLF4J bundle is activated first (or used by another bundle without being activated).
2. The Logback bundle is activated only later.

Since `LocationAwareLogger` is the oldest available mechanism to transfer caller boundary information, it is commonly used in API-to-API bridges, e.g. the [Log4j API-to-SLF4J bridge](https://github.com/apache/logging-log4j2/blob/e749fa135f78dcdf5952fe51c8191a57250c1614/log4j-to-slf4j/src/main/java/org/apache/logging/slf4j/SLF4JLogger.java#L297-L328) or [Spring JCL-to-SLF4J bridge](https://github.com/spring-projects/spring-framework/blob/d8c153a9d118ca508dccff846b14b672137d4d47/spring-jcl/src/main/java/org/apache/commons/logging/LogAdapter.java#L403).

Newer mechanisms such as `LoggingEventAware` (for logger implementations) or `CallerBoundaryAware` (for logger builder implementations) are less common. The default `LoggingEventBuilder` also requires the underlying logger to either implement `LocationAwareLogger` or `LoggingEventAware`:

https://github.com/qos-ch/slf4j/blob/69c333de280100f7dc99ee00e302192690fcc761/slf4j-api/src/main/java/org/slf4j/spi/DefaultLoggingEventBuilder.java#L157-L169

Theschorpioen (approved):

PR #439, opened 2024-10-28
Fixes location information in `JDK14LoggerAdapter`

ppkarwasz:
The `JDK14LoggerAdapter` class uses the wrong caller boundary for location unaware methods.

This PR:

- Sets the correct caller boundary for classic API method calls (`AbstractLogger`),
- Adds the missing `LoggingEventAware` interface and fixes its implementation. The interface was previously implemented, but not declared.
- Simplifies the fix for #425 and adds a test for that issue.

The `SubstituteLogger` tests depend on #438.


ppkarwasz:
I am converting this to draft, since it requires #438.

Theschorpioen (approved):

firoj0:
#438

PR #443, opened 2024-11-29
Fix JavaDoc of MDCAmbit.java

henryju:


Theschorpioen (approved):

ceki:
@henryju Thank you for catching this. Please keep in mind that PRs need to be signed-off, i.e. the -s flag must be set when issuing the "git commit" command.

henryju:
I submitted the change using GitHub online editor directly, not sure how I can sign in this case.

ceki:
@henryju I see. I had not considered the online editor. Thank you for the explanation.

PR #456, opened 2025-04-03
Fix trace-level logs showing up as DEBUG due to FINER mapping

sglee487:
While lowering the log level to TRACE to monitor more detailed logs, I noticed that some logs that should be TRACE were actually showing up as DEBUG.

I found that jul’s FINER level was being mapped to DEBUG, so I updated the mapping so that FINER is treated as TRACE instead, and also fixed the related tests.


ceki:
Hello @sglee487, 

Thank you for this PR.

The mapping from java.util.logging levels to SLF4J level, more specifically from FINEST to TRACE, FINER to DEBUG and FINE to DEBUG is documented in the [SLF4JBridgeHandler javadocs](https://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html). I am afraid any change in the above mapping could break existing deployments. As such, I am closing this issue as WONT_FIX.

PR #462, opened 2025-05-21
add support for peekByKey in MDCAdapter

### LINKED ISSUE #461: add support for peekByKey in MDC
In some scenarios, after using `MDC.pushByKey(key, value)` to **push** a value into the deque (stack), we may need to retrieve the top value without **popping** the stack referenced by 'key', similar to a ​​`peekByKey​​` operation. However, the current SLF4J implementation requires developers to invoke `getCopyOfDequeByKey(key)` and manually **peek** from the returned deque (stack), which involves creating a copy of the data structure and leads to unnecessary performance overhead. Would it be more efficient to add native support for `peekByKey` in MDC to avoid this redundancy?

Thank you for your attention and support.

youjie23:
fixed #461 add support for peekByKey in MDCAdapter
1、add `peekByKey` in MDCAdapter with default implement
2、add `peekByKey` in  `ThreadLocalMapOfStacks` and `BasicMDCAdapter`


ppkarwasz (approved):
These changes look good to me, so I’d like to express my **(non-binding)** support for including them in SLF4J.

The absence of a `peekByKey` method in `ThreadLocalMapOfStacks` was the primary reason I had to reimplement this functionality in `log4j-slf4j2-impl`, so this addition is particularly welcome.

One question I have is about versioning: how are these changes intended to be adopted by implementors? As I understand it, SLF4J 2.0 implementations are expected to maintain compatibility with all 2.x versions starting from `2.0.0`. Would it be more appropriate for these changes to be introduced in `2.1.0`, so that only implementations explicitly targeting `2.1.x` can use them?

youjie23:
Thanks everyone for the feedback and review on this pull request, especially @ppkarwasz.

This pull request has accumulated a large number of historical synchronization commits, which makes the actual change difficult to review. I have therefore moved the peekByKey implementation into a clean branch based on the latest master.

The new pull request is:
#492  — Add non-destructive MDCAdapter peekByKey for issue #461

The new PR contains only the relevant API, implementation, documentation, and tests. It also addresses the versioning concern raised in the review by targeting the current SLF4J 3.0 development line.

I am closing this pull request to avoid duplicating review effort. If possible, I would appreciate it if the previous reviewers could continue the review on the new PR.

Issue #472, opened 2025-11-18
Newline character is configurable


vitalijr2:
Hi guys,

some years ago I have found an AWS CloudWatch issue with truncating logs if the a logger uses a standard `\n` character.

Some discussions and related issues:
- https://github.com/Graylog2/graylog-plugin-aws/issues/159
- [How to Keep Multiline Logs as Single Records in CloudWatch with Node.js](https://itnext.io/how-to-keep-multiline-logs-as-single-records-in-cloudwatch-with-node-js-8a9285c83800)
- [[Does Amazon CloudWatch Logs support a newline character?](https://stackoverflow.com/questions/43009755/does-amazon-cloudwatch-logs-support-a-newline-character)](https://stackoverflow.com/questions/43009755/does-amazon-cloudwatch-logs-support-a-newline-character)

There is a small fix for AWS Lambdas: use `\r` instead of `\n`.

How about to add this feature to SimpleLogger: [write to an output stream logging message + configurable ending](https://github.com/qos-ch/slf4j/blob/32a9854cf414993abcefec097354ca1758dc45ec/slf4j-simple/src/main/java/org/slf4j/simple/SimpleLogger.java#L261)?

ceki:
@vitalijr2 Hi! I'm not entirely sure about the context of this issue yet. Feel free to fork the project – I'd be happy to link to your fork from the official SLF4J documentation and website. We can take it and figure things out from there.

PR #473, opened 2025-11-19
CloudWatch Newline Fix

vitalijr2:
As I wrote in #472 there is the specific **AWS CloudWatch** issue related to using the linefeed character `\n` as a separator.

And the quick fix exists: replace `\n` with `\r`.

I have added a new configuration key **cloudwatchNewlineFix**: when it is `true` all linefeed characters are replaced with carriage return ones.

I prefer the name **cloudwatchNewlineFix** instead of **replaceNewlineWithCarriageReturn** as it is a specific feature and should not be used on any other platform.


ceki:
@vitalijr2 Thank you for this PR. As mentioned earlier, I am unfamiliar with CloudWatch. Thus, I suggest that you publish this version of `SimpleLogger` as a separate artifact under an groupId:artifactId of your choice. How does that sound?

Issue #464, opened 2025-05-23
avoid logViaPublicSLF4JLoggerAPI showing with log messages


pruy:
Version 2.0.17
When using the fluent API (via atXXXX() methods on a Logger instance)
the resulting log message always shows "logViaPublicSLF4JLoggerAPI
as the calling method for %M message format token.
This is rather inconvenient when trying to get the method name with the log message.

I'd expect such functionality not showing an "internal" method.

AzazelSensei:
Opened https://github.com/qos-ch/slf4j/pull/488 — current master already reports the caller class; the test now pins the method name too.

ceki:
@pruy  Thank you for the report.

With `slf4j-jdk14`, fluent logging (`atXxx()` / `DefaultLoggingEventBuilder`) should go through `logViaLocationAwareLoggerAPI`, not `logViaPublicSLF4JLoggerAPI`. The latter is used only when the logger instance implements neither `LoggingEventAware` nor `LocationAwareLogger`.

`JDK14LoggerAdapter` does implement `LocationAwareLogger`, so if `%M` shows `logViaPublicSLF4JLoggerAPI`, the fluent API is probably not talking to that adapter directly. Typical causes:

- a wrapper `Logger` that only implements the public SLF4J `Logger` interface
- another facade or adapter on the classpath (for example a JCL/Log4j/JUL bridge, or a library wrapper)

Could you list the logging-related jars on the classpath (in particular any `*-over-slf4j`, `slf4j-to-*`, or custom logger wrappers) and, if possible, a small reproducer?

pruy:
I have put together an illustrating example.
Extract zip file
Build using `mvn package`
Call using `java -jar target/issue464-0.0.1-SNAPSHOT-jar-with-dependencies.jar`

All output lines are expected to show the proper method name. But not all do.

[Slf4jIssue464.zip](https://github.com/user-attachments/files/31796546/Slf4jIssue464.zip)

ceki:
Thank you for this reproducer. The issue is specific to XLogger part of slf4j-ext. 

Here is a diagnosis of the issue.

The fluent API (`atInfo().log(...)` and the other `atXxx()` methods) does not go through `Logger.info()` / `debug()` / etc. It goes through `Logger.makeLoggingEventBuilder(Level)`.

The default implementation of that method, on the `Logger` interface, is:

```java
default LoggingEventBuilder makeLoggingEventBuilder(Level level) {
[...]

PR #488, opened 2026-08-13
Pin fluent API caller method name in CallerInfoTest

AzazelSensei:
On current master, logger.atDebug().log already reports the test class, not logViaPublicSLF4JLoggerAPI. I added an assert on the method name so %M cannot regress to the internal helper.

See #464


AzazelSensei:
Still nothing. Closing. Shame, it was a small fix.

ceki:
Reopening for merge.

